PR #27, opened 2014-12-05
openstreemap pipeline improvements

missinglink:
PENDING REVIEW

First of all; apologies for the very large PR. This work is the result of research and development of the best way to build the OSM import pipeline over the last 3-4 months.

The result is a far more refined, stable and tested import pipeline which we can continue to build off going-forward.

The current master branch (the target of this PR) has 4 major pain points:

1. `slow imports` - it can take up to 20 days to import the full planet file, it's not multi-core.
2. `failure to exit(0)` - sometimes the imports fail to exit cleanly, making ops scripting difficult.
3. ` Recursive process.nextTick detected` - streams-related errors.
4. `code clarity/testing` - the code is only partially tested and in some places, it's messy. 

In order to address [1] the following work was done, which resulted in drastic speed improvements:

- [x] implement `golang` pbf parser 
- [x] implement admin-lookup
- [x] upgrade to the latest suggester-pipeline module
- [x] simplify features.js
- [x] remove osm_types.js

In order to fix [2], I simply:

- [x] fix stats.js module via unit tests, remove complexity

Issue [3] was resolved by fixing the above issues and:

- [x] upgrade all dependencies.
- [x] fix nodejs version inconsistencies to work with both `0.10` and `0.12`.

Issue [4] was tackled by:

- [x] implement pelias-model
- [x] add doc.setAddress() and doc.getAddress() functions to pelias/model
- [x] general clean up
- [x] improve global module.exports, add tests.
- [x] configure code linting and precommit hook, lint everything.
- [x] make osm data mappers clearer and easier to write/modify
- [x] add a unit test for every module
- [x] add detailed comments at the top of complex streams to explain their purpose
- [x] add end-to-end system test

Some extra benefits we got along the way:

- [x] extract address data and store in ES
- [x] remove dependency on having quattroshapes indexed before running
- [x] resolve inconsistent ways count from subsequent import runs


sevko:
Before this gets merged, I'm curious how you identified the OSM parsing as our bottleneck.

missinglink:
In the case of OSM data its pretty obviously a bottleneck as we parse ~2.5B nodes and only use ~20m of them, so the faster it runs the better off we will be.

Having said that I agree the other I/O operations still need better performance testing.

sevko:
Right, but I can parse an 800mb PBF in 3 minutes with straight up `osm-pbf-parser`. If we assumed that the rest of our pipeline ran instantaneously, wouldn't the entire import take that long? Since it doesn't, it seems like the delays are elsewhere. Am I missing something?

missinglink:
That is a very good question.

sevko:
Let's figure that out and discuss how we should benchmark the importer before this gets merged.

cc @dianashk , @hkrishna

sevko:
Well, it turns out that I can parse *and* filter the planet PBF with pure JavaScript (`osm-pbf-parser` and this importer's `osm_filter` stream) in ~17 hours. Still don't understand why the Go parser is making a significant difference.

heffergm:
My 2p, solely based on my observations running both, are: 
- with old setup, no activity on multiple cores (i.e. single CPU @100%) regardless of thread count, peak of 4k requests/s to ES
- with experimental setup, activity on all cores, peak of ~30k res/s to ES

If I had to hazard a guess, there's a limitation in the old implementation that's blocking on request/response to ES, not necessarily parsing the pbf, and the fact that it's operating only on a single core effectively limits the throughput you're going to get.

sevko:
Is this ready to get merged?

dianashk:
@missinglink is currently investigating this issue:

```
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.

RangeError: Maximum call stack size exceeded
```

not ready to merge

sevko:
Indeed, I just ran into that in some local imports. Running with node 0.12.0 seems to fix it.

missinglink:
related: https://github.com/pelias/pelias/issues/56

missinglink:
@pelias/owners ready for review & merge

sevko:
@missinglink , awesome work. Some observations:

* The `util/geoJsonCenter` and `util/geoJsonType` modules seem like they should either be moved into `stream/denormalizer`, or at least combined into one module (having only one function per module hinders project navigability).

* I think you should just make the files in `schema/` one JSON file, since they're all fairly concise.

* `stream/address_extractor` appears to extract admin values. I was under the impression that we decided to avoid doing that in our importers (in `pelias-geonames`, anyway) to ensure admin-value consistency.

index.js
```diff
@@ -1,253 +1,37 @@
...
-  any: {
-    hierarchyLookup:            require('./stream/osm/any/hierarchyLookup')
-  }
+var elasticsearch = require('pelias-dbclient'),
+    adminLookup = require('pelias-admin-lookup'),
+    suggester = require('pelias-suggester-pipeline'),
+    dbmapper = require('./stream/dbmapper');
+
+var osm = { pbf: {}, doc: {}, address: {}, tag: {} };
+exports = module.exports = osm;
```
sevko: Why is that `exports =` at the beginning of the line? Seems like an extraneous assignment.

index.js
```diff
@@ -1,253 +1,37 @@
...
-  },
-  any: {
-    hierarchyLookup:            require('./stream/osm/any/hierarchyLookup')
-  }
+var elasticsearch = require('pelias-dbclient'),
+    adminLookup = require('pelias-admin-lookup'),
+    suggester = require('pelias-suggester-pipeline'),
+    dbmapper = require('./stream/dbmapper');
+
+var osm = { pbf: {}, doc: {}, address: {}, tag: {} };
```
sevko: I think it'd be cleaner if you replaced the `osm` object with simple variable declarations, ie:

```javascript
var docConstructor = require('./stream/document_constructor');
var docDenormalizer = require('./stream/denormalizer');
```

It doesn't really seem to serve any purpose.

missinglink:
thanks for the review @sevko 

1. I was hoping we could remove that functionality completely, are you working on something like that?
2. I don't see the value of merging these, there is one schema for names and one for addresses, I assume you mean merging the addresses one? these are already merged for performance reasons here https://github.com/pelias/openstreetmap/blob/experimental/stream/tag_mapper.js#L11 but left seperate for readability.
3. It does not extract admin values, it simply copies them from the poi doc to the address doc so we don't have to look them up twice.

index.js
missinglink: osm is exported. the object hierarchy is intended to give readability to the exported API

index.js
missinglink: I tend to agree, being too hispter here

sevko:
1. Nope. We still need to compute way center-points until we decide to import non-point geometries into elasticsearch, so it seems like those should stick around.
2. They hinder navigability more than improve readability, in my opinion. Two of those files have one key-value pair, and a lot of boilerplate (including a `module.exports`, which a `.json` file wouldn't need, and the same header comment duplicated in four places).
3. My mistake.

missinglink:
1. I think we should aim for a separate module as this functionality may be useful to other modules in the future, the code itself is covered by a stream test but the libs not unit tested. I personally feel it's out of scope of this PR and would add more complexity to an already large PR which can be merged independently of that issue.

2. What does `hinder navigability` mean? If you are saying that code with less characters is better then I strongly disagree. A json file cannot contain comments and would be far more confusing to other developers trying to collaborate.

sevko:
I think that if we want to make any more changes, now's as good a time as ever.

1. In that case, at least compress them into one module. One function per module seems extreme.
2. I mean navigating three files is more difficult than navigating one, especially when they contain hardly any content. Feel free to keep them as a `.js` file (indeed, losing the comment would be bad), but I don't see why each deserves its own module.

missinglink:
In regards to abstraction, modularity, interface segregation and readability we can probably agree to disagree. I don't see us getting anywhere on such subjective topics.

Do you feel strongly enough about these issues that we should not merge this PR in it's current state?

sevko:
The easier a source tree is to navigate and read, the more likely we are to get external contributors. This seems like it'll confuse people (why so many files/directories for something so simple?). It did me, at least.. @dianashk , what's your take?

index.js
dianashk: in my opinion this would work better as a class, especially since it's being exported.

util/geoJsonCenter.js
```diff
@@ -16,4 +20,4 @@ module.exports = function( geometry ){
   }
 
   return;
```
dianashk: can we return false or null here?

stream/pbf.js
```diff
@@ -0,0 +1,68 @@
...
+  returning a valid readable stream.
+**/
+
+var fs = require('fs'),
+    pbf2json = require('pbf2json'),
+    settings = require('pelias-config').generate(),
+    features = require('../config/features');
+
+// Create a new pbf parser stream
+function createPbfStream( opts ){
```
dianashk: please document what opts is expected to have, like

/**
 * @param {object} opts
 * @param {string} opts.file ...
 * @param {string} opts.leveldb ...
 * etc...
 */

stream/document_constructor.js
```diff
@@ -0,0 +1,42 @@
...
+    doc.setMeta( 'tags', item.tags || {} );
+
+    // Push instance of Document downstream
+    this.push( doc );
+    return next();
+
+  });
+  
+  // catch stream errors
+  stream.on( 'error', console.error.bind( console, __filename ) );
```
dianashk: should we log to pelias-logger instead of console?

dianashk:
To address the earlier discussion:

1. With regard to combining/moving the geoJson*.js util contents, I find the current split clean and modular. I can see the reasoning for splitting little utilities like that up into separate files. Grouping them under the util directory hides the perceived complexity of too many files, while having them in separate files means you don't have to scroll through a wall of code to find a small function not at all related to the rest of the file contents. It's not unlike node_modules/ hiding all the various dependencies.

2. For the same reason as (1), I don't mind the schema breakup. There is a logical split between those schema items, and having them split up in separate files highlights that to the reader. I'm not offended by this. What could be done to make it clear that the various address parts should be merged and to hide that complexity, is to add an index.js to that directory. The require the schema directory instead of individual files.

`schema/index.js`

```javascript
module.exports.NAME_SCHEMA = require('../schema/name_osm');
module.exports.ADDRESS_SCHEMA = merge( true, false,
  require('./address_osm'),
  require('./address_naptan'),
  require('./address_karlsruhe')
);
```

`stream\tag_mapper.js`

```javascript
var schema = require('../schema');
```

stream/document_constructor.js
dianashk: I see switching to pelias-logger is a larger undertaking than this bloated PR should have to take. :) I will open a separate issue.

sevko:
One function per file still seems foreign to me, since modules are supposed to be logical groupings of like functions. More importantly, with regards to the schema files, duplicating the *same* header comment in three places and having a file that looks like:

```javascript

/**
  Attempt to map OSM address tags to the Pelias address schema.
  On the left is the OSM tag name, on the right is corresponding
  doc.address key for which it should be mapped to.
  eg. tags['naptan:Street'] -> doc.address['street']
  @ref: http://wiki.openstreetmap.org/wiki/NaPTAN
**/

var NAPTAN_SCHEMA = {
  'naptan:Street': 'street'
};

module.exports = NAPTAN_SCHEMA;
```

just seems smelly. At this point, though, two people disagree and this is relatively unimportant so I don't have much of an argument. Before I +1, have we run a staging (ie full planet) import with the latest commit in this branch?

index.js
```diff
@@ -1,253 +1,38 @@
...
+  The openstreetmap importer provides a way of parsing, mapping and
+  augmenting OSM data in to elasticsearch.
+**/
 
-var backend = function( index, type ){
-  return new Backend( dbclient.client, index, type );
+var elasticsearch = require('pelias-dbclient'),
+    adminLookup = require('pelias-admin-lookup'),
+    suggester = require('pelias-suggester-pipeline'),
+    dbmapper = require('./stream/dbmapper');
```
dianashk: why are all the stream object defined in this project required and saved as part of the osm object except dbmapper? would we expect the client of this module to want access to all the other objects but not dbmapper, or adminLookup/suggester for that matter? just curious what the distinction is here.

test/stream/stats.js
```diff
@@ -5,92 +5,47 @@ var stats = require('../../stream/stats'),
 module.exports.tests = {};
```
dianashk: I don't see stats being used in the code anymore (other than tests). Can it be removed altogether?

dianashk:
Overall, what a huge difference! Definitely a lot more streamlined and polished. Tests are a great addition. :+1:

index.js
missinglink: why? could you make a case please?

util/geoJsonCenter.js
missinglink: `return;` is equivalent to `return undefined;`, why are `return false;` or `return null;` better?

index.js
missinglink: There are pros/cons to this approach, the `dbmapper` is a pretty generic stream which is used to map `pelias.model.Document` objects to the syntax that is accepted by `pelias/dbclient`.

Exporting it makes it easier for 3rd party consumers to do that simple mapping **but** it strictly targets the `pelias` index, meaning consumers cannot build an index of another name.

I would love to see this functionality abstracted out to either `pelias/dbclient`, `pelias/model` or another module which is responsible for this domain and re-usable by other imports, for that reason I chose not to export the stream in order to avoid deprecating that API in the future.

test/stream/stats.js
missinglink: It could be moved to an external module, I will open a ticket for this and the other tasks to externalise functionality not in the osm domain.

stream/pbf.js
missinglink: I thought we agreed to discuss this as a team?

stream/document_constructor.js
missinglink: Please do.

index.js
missinglink: see: https://github.com/pelias/openstreetmap/commit/9d4edc379de3f65c30e7a035ea1a8c8babb850d5

index.js
dianashk: When you export a class you can be a lot more explicit about the interface. Do things like define accessor methods if needed. Allow multiple instances, if that's a desired behavior, and explicitly indicate it isn't by making it a singleton. Maybe it's my OOP background, but exporting an anonymous object as the face of your package feels unfinished. With that said, I'm not extremely passionate about this and will not be super upset if you keep it as is.

index.js
```diff
@@ -1,253 +1,38 @@
...
 
-// @todo: extract this or refactor suggester to make it a stream factory
-var bun = require('bun');
-function generateSuggester(){
-  return bun([
-    suggester.streams.suggestable(),
-    suggester.streams.suggester( suggester.generators )
-  ]);
+// run import if executed directly; but not if imported via require()
+if( require.main === module ){
```
dianashk: I missed it yesterday, but I think it's important. `index.js` should be used strictly for exporting your package and shouldn't be directly executed. There should be an additional `app.js` / `server.js` file that requires the `index.js` and executes. That entry point then also serves as an example of usage for clients. I think this is the common expectation when looking at a node project and is based on the following. `index.js` is the default file used when `require`ing a directory, hence `index.js` makes sense as the main export point for a module. `npm start` will default to `node server.js` if a `server.js` file is found in the package root. So that means a separate entry point for execution is expected.)

util/geoJsonCenter.js
dianashk: returning `undefined` means the function is of type `void` which it isn't. if there is ever an expectation of the function returning something, it shouldn't return void at any point. `null` would be an appropriate substitution for an object when one cannot be created.

stream/pbf.js
dianashk: Sure, we can discuss later. But I made the comment in this place specifically because opts is obscuring the interface and even if we don't decide to add doc blocks across the board, a comment here about what opts is expected to have is something I think is sorely missing.

util/geoJsonCenter.js
sevko: Agreed. It's a semantic difference. `null` implies a result could not be computed, while `undefined` implies a variable wasn't initialized.

index.js
dianashk: Thanks for the explanation.

test/stream/stats.js
dianashk: Great.

index.js
sevko: I think the database mapper stream should be moved into `pelias-dbclient`, since we're currently repeating that same bit of code in all of our importers. If we're still keen on keeping that package generalized for anyone's use, we should then re-publish it as an `elasticsearch-bulk-indexer` and rewrite `pelias-dbclient` as a lightweight wrapper for it.

stream/pbf.js
sevko: Indeed, doesn't have to be in JSDoc format, but an enumeration of all the properties would be great.

index.js
missinglink: A 'class' in JS represents an instantiable construct for which each instance may hold state for it's individual properties while sharing methods from a prototype.

This functionality is not required in this case and so we would end up exporting a singleton with no individual state or prototypal methods.

The current implementation is equivalent and forwards compatible if (hypothetically) we decided to do export a singleton in the future.

My biggest concern here is a fundamental misunderstanding of the architecture, the system is designed in a way that allows individual components to be easily added/removed from the import pipeline by simply excluding/including additional streams to the pipeline.

A good analogy is this:

```
#! /bin/bash
$> osm_parser | admin_lookup | elasticsearch;
```

if another developer would like to decompose that pipeline and fork an existing stream or add a new one then they can simply re-compose the pipeline as such:

 ```
#! /bin/bash
$> osm_parser | my_fork_of_admin_lookup | something_else | elasticsearch;
```

If we move towards a more traditional OOP architecture then we would potentially lose the advantages we gained from the unix pipes style interfaces in that interoperability and composability are sacrificed in favour of readability for traditional OOP developers.

As it currently stands this repository is completely functional and stateless, I would prefer to keep it that way.

I think this specific discussion would be best continued outside of this PR.

File: index.js (hunk 8 above)
Comment (missinglink) [inline index.js]: This is a fair comment but unfortunately it will need to stay this way until we can update our dependants in `pelias/vagrant` and our build scripts as they currently rely on executing the import via `node index.js`.

**Comment by dianashk** [comment]
+1

**Comment by sevko** [comment]
@missinglink ,

> Before I +1, have we run a staging (ie full planet) import with the latest commit(s) in this branch? I don't remember whether things like the admin-lookup integration were included in our last run.

File: util/geoJsonCenter.js (hunk 3 above)
Comment (missinglink) [inline util/geoJsonCenter.js]: Are we talking about javascript here or another language, how is this function type 'void'?

If we are talking `types` in js, then exporting something which is an object seems a bit odd to me.
re: `typeof null == "object"`.

Since you feel strongly about this, I am happy to change to accommodate.

**Comment by missinglink** [comment]
It's a pity this PR turned out to be a chance for everyone to voice their personal opinion about their individual programming style.

**Comment by sevko** [comment]
Well, code reviews are meant to cover changes that would result in objectively cleaner/better code in addition to purely functional improvements. We all clearly have strong opinions about the former, and (dis)agree about a number of them. I think it's good that we got a chance to discuss them.

**Comment by missinglink** [comment]
@heffergm can you please test this in our dev environment. experimental branch is now merged to master.

```bash
$ npm publish
npm http 201 https://registry.npmjs.org/pelias-openstreetmap
+ pelias-openstreetmap@1.0.0
```

sevko:
In the future, should we test branches in imports before merging them to master? I did that with both [pelias-geonames](https://github.com/pelias/geonames/pull/19) and [https://github.com/pelias/quattroshapes-pipeline](https://github.com/pelias/quattroshapes-pipeline/pull/17). Seems like that's the best way to guarantee that only stable code hits `master`.

missinglink:
@sevko
- This branch has been tested numerous times in development.
- It is covered by both unit tests and end-to-end tests.
- We do not have a staging environment.
- A full planet build takes ~6 days.

PR #52, opened 2015-04-13
Error handling

missinglink:
A bunch of improvements regarding error handling:

- gracefully handle newline and tab characters in OSM tag values
- catch generic thrown errors in stream transform functions

Plus some minor tweaks:

- rename `item` to `doc` for consistency ($doc represents a `pelias/model.Document` object)
- rename `done` to `next`, again for consistency
- flatten some nested code to reduce cyclomatic complexity
- add more tests to cover more edge-cases

### DIFF
--- a/stream/address_extractor.js
+++ b/stream/address_extractor.js
@@ -88,6 +88,7 @@ module.exports = function(){
     catch( e ){
       console.error( 'address_extractor error' );
       console.error( e.stack );
+      console.error( JSON.stringify( doc, null, 2 ) );
     }
 
     return next();
--- a/stream/address_extractor.js
+++ b/stream/address_extractor.js
@@ -29,59 +29,69 @@ var through = require('through2'),
     Document = require('pelias-model').Document,
     idOrdinal = 0; // used for addresses lacking an id (to keep them unique)
 
-function hasValidAddress( item ){
-  if( !isObject( item ) ){ return false; }
-  if( !isObject( item.address ) ){ return false; }
-  if( 'string' !== typeof item.address.number ){ return false; }
-  if( 'string' !== typeof item.address.street ){ return false; }
-  if( !item.address.number.length ){ return false; }
-  if( !item.address.street.length ){ return false; }
+function hasValidAddress( doc ){
+  if( !isObject( doc ) ){ return false; }
+  if( !isObject( doc.address ) ){ return false; }
+  if( 'string' !== typeof doc.address.number ){ return false; }
+  if( 'string' !== typeof doc.address.street ){ return false; }
+  if( !doc.address.number.length ){ return false; }
+  if( !doc.address.street.length ){ return false; }
   return true;
 }
 
 module.exports = function(){
 
-  var stream = through.obj( function( item, enc, done ) {
+  var stream = through.obj( function( doc, enc, next ) {
 
-    var isNamedPoi = !!item.getName('default');
+    try {
 
-    // create a new record for street addresses
-    if( hasValidAddress( item ) ){
-      var type = isNamedPoi ? 'poi-address' : 'address';
+      var isNamedPoi = !!doc.getName('default');
 
-      // copy data to new document
-      var record = new Document( 'osmaddress', type + '-' + item.getType() + '-' + (item.getId() || ++idOrdinal) );
+      // create a new record for street addresses
+      if( hasValidAddress( doc ) ){
+        var type = isNamedPoi ? 'poi-address' : 'address';
 
-      record
-        .setName( 'default', item.address.number + ' ' + item.address.street )
-        .setCentroid( item.getCentroid() );
+        // copy data to new document
+        var record = new Document( 'osmaddress', type + '-' + doc.getType() + '-' + (doc.getId() || ++idOrdinal) );
 
-      // copy address info
-      record.address = item.address;
+        record
+          .setName( 'default', doc.address.number + ' ' + doc.address.street )
+          .setCentroid( doc.getCentroid() );
 
-      // copy admin data
-      if( item.alpha3 ){ record.setAlpha3( item.alpha3 ); }
-      if( item.admin0 ){ record.setAdmin( 'admin0', item.admin0 ); }
-      if( item.admin1 ){ record.setAdmin( 'admin1', item.admin1 ); }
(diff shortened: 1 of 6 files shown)


stream/address_extractor.js
```diff
@@ -29,59 +29,69 @@ var through = require('through2'),
...
-      this.push( record );
     }
 
-    // forward item downstream is it's a POI in it's own right
-    // note: this MUST be below the address push()
-    if( isNamedPoi ){
-      this.push( item );
+    catch( e ){
+      console.error( 'address_extractor error' );
+      console.error( e.stack );
```
dianashk: would it be helpful to output the doc here as well? so it's easier to debug issues.

stream/address_extractor.js
```diff
@@ -29,59 +29,69 @@ var through = require('through2'),
...
-  if( !isObject( item ) ){ return false; }
-  if( !isObject( item.address ) ){ return false; }
-  if( 'string' !== typeof item.address.number ){ return false; }
-  if( 'string' !== typeof item.address.street ){ return false; }
-  if( !item.address.number.length ){ return false; }
-  if( !item.address.street.length ){ return false; }
+function hasValidAddress( doc ){
+  if( !isObject( doc ) ){ return false; }
+  if( !isObject( doc.address ) ){ return false; }
+  if( 'string' !== typeof doc.address.number ){ return false; }
```
sevko: It might be cleaner to rewrite the body of this function as a single conditional, like:

```javascript
return isObject(doc) && isObject(doc.address) /* etc */
```

stream/address_extractor.js
missinglink: Good call!

https://github.com/pelias/openstreetmap/commit/8e5e0afdfbc1c2a02806c66864800ebe524a5096

```bash
category_mapper error
Error: test
    at Document.doc.getMeta (/var/www/pelias/openstreetmap/test/stream/category_mapper.js:178:37)
    at DestroyableTransform._transform (/var/www/pelias/openstreetmap/stream/category_mapper.js:30:22)
    at DestroyableTransform.Transform._read (/var/www/pelias/openstreetmap/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:184:10)
    at DestroyableTransform.Transform._write (/var/www/pelias/openstreetmap/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:172:12)
    at doWrite (/var/www/pelias/openstreetmap/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:237:10)
    at writeOrBuffer (/var/www/pelias/openstreetmap/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:227:5)
    at DestroyableTransform.Writable.write (/var/www/pelias/openstreetmap/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:194:11)
    at Test.<anonymous> (/var/www/pelias/openstreetmap/test/stream/category_mapper.js:186:12)
    at Test.bound [as _cb] (/var/www/pelias/openstreetmap/node_modules/tape/lib/test.js:62:32)
    at Test.run (/var/www/pelias/openstreetmap/node_modules/tape/lib/test.js:75:10)
{
  "name": {},
  "address": {},
  "center_point": {},
  "category": []
}
```

dianashk:
:+1:

stream/address_extractor.js
```diff
@@ -29,59 +29,70 @@ var through = require('through2'),
...
-      if( item.admin1_abbr ){ record.setAdmin( 'admin1_abbr', item.admin1_abbr ); }
-      if( item.admin2 ){ record.setAdmin( 'admin2', item.admin2 ); }
-      if( item.local_admin ){ record.setAdmin( 'local_admin', item.local_admin ); }
-      if( item.locality ){ record.setAdmin( 'locality', item.locality ); }
-      if( item.neighborhood ){ record.setAdmin( 'neighborhood', item.neighborhood ); }
-      
-      // copy meta data (but maintain the id & type assigned above)
-      record._meta = extend( true, {}, item._meta, { id: record.getId(), type: record.getType() } );
+        // copy address info
+        record.address = doc.address;
```
sevko: We should probably use the Document setters here, in case their internal `.address` schema changes.

stream/address_extractor.js
```diff
@@ -29,59 +29,70 @@ var through = require('through2'),
...
+
+        // copy admin data
+        if( doc.alpha3 ){ record.setAlpha3( doc.alpha3 ); }
```
sevko: If any of these setters throw an exception, the call stack will unwind all the way up to the `try {` on like 46 and thus discard this record entirely. Since they're not 100% necessary (just nice to have) and we can push records down the pipeline without them, they should probably be wrapped in their own `try`/`catch`es. Also, the scope of the `try`/`catch` block that begins on line 46 could be narrowed down a bit, say down to line 55 when you call the first Document functions that you might expect to throw exceptions? Otherwise, you might be unintentionally suppressing unrelated errors.

sevko: To illustrate, [here's how pelias/geonames handles it](https://github.com/pelias/geonames/blob/master/task/import.js#L14,L54).

sevko: Basically, I'm envisoning something like:

```javascript
var record = new Document( 'osmaddress', type + '-' + doc.getType() + '-' + (doc.getId() || ++idOrdinal) )
  .setName( 'default', doc.address.number + ' ' + doc.address.street )
  .setCentroid( doc.getCentroid() );

// copy address info
record.address = doc.address;

// copy admin data
if( doc.alpha3 )
  try {
    record.setAlpha3( doc.alpha3 );
  } catch( ex ){}
}

if( doc.admin0 ){
  try {
    record.setAdmin( 'admin0', doc.admin0 );
  } catch( ex ){}
}
```

missinglink: I don't think it's possible for those setters to throw as they get their data from a previously set model method which has already validated the data?

sevko: Oh, I see. Can you use the `Document` getters instead of accessing properties directly? I made a similar comment about using setter methods [here](https://github.com/pelias/openstreetmap/blob/error_handling/stream/address_extractor.js#L62). That way, things won't break if the internal `pelias-model` schema changes.

missinglink:
```bash
$ npm publish
+ pelias-openstreetmap@1.4.0
```

PR #57, opened 2015-04-14
Use document getters#56

### LINKED ISSUE #56: use Document getters
As per the discussion in #52 and #53, [this code](https://github.com/pelias/openstreetmap/blob/master/stream/address_extractor.js#L65,L72) should use `Document` getter methods if the `doc` variable is itself a `Document` instead of accessing properties directly (so, `doc.getAlpha3()` instead of `doc.alpha3`).

sevko:
Some modifications to `stream/address_extractor.js`: reduce the scope of the overarching `try`/`catch` block and use `Document` getters/setters.

Fixes #56

### DIFF
--- a/stream/address_extractor.js
+++ b/stream/address_extractor.js
@@ -55,18 +55,24 @@ module.exports = function(){
           .setName( 'default', doc.address.number + ' ' + doc.address.street )
           .setCentroid( doc.getCentroid() );
 
-        // copy address info
-        record.address = doc.address;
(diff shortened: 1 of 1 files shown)


dianashk:
:+1:

missinglink:
the try/catch blocks are not required when using `set(get())` on data which has previously been validated. other than that LGTM.

sevko:
The values returned by `.get*()` could be `undefined` if they were never set.

missinglink:
oh yea, good call, :+1:

PR #61, opened 2015-05-04
larger fixture file

missinglink:
upgrade to a larger fixture file in order to catch a wider range of regression errors


dianashk:
:+1:

missinglink:
I'm going to change the fixture before merging this, this Vancouver extract doesn't seem to have a single category in the whole file!

```bash
node -e "var j=require('./test/fixtures/vancouver.extract.expected'); Object.keys(j).forEach(function(k){var i=j[k]; var c=i.data.category; if(c.length) console.log(c); });"
```

Issue #21, opened 2014-10-28
oddities in addr:housenumber tag


missinglink:
It seems some house numbers are actually lists of house numbers.

In this case it might be best to create multiple records for these entries due to the suggester being anchored on the left side of the input.

These records **may** also include interpolation information, which is worth looking in to.

ref: http://wiki.openstreetmap.org/wiki/Addresses#Multiple_buildings_for_one_housenumber
"Some mappers use addr:interpolation=* on a single object to indicate that its addr:housenumber=* is to be treated as a range rather than an atomic label (discussed above). There is not a consensus on whether this is appropriate."

eg: http://pelias.mapzen.com/doc?id=osmway:poi-address-way-236566013

```javascript
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          -1.456305,
          50.919062
        ]
      },
      "properties": {
        "name": "1;2;3;4;5;6;7;8;9;10 Pennine Road",
        "alpha3": "GBR",
        "admin0": "United Kingdom",
        "admin1": "City Of Southampton",
        "admin2": "Hampshire",
        "locality": "Southampton",
        "neighborhood": "Wimpson",
        "text": "1;2;3;4;5;6;7;8;9;10 Pennine Road, Wimpson, City Of Southampton"
      }
    }
  ],
  "date": 1414495445297
}
```

missinglink:
ref: http://wiki.openstreetmap.org/wiki/Talk:Semi-colon_value_separator

PR #63, opened 2015-05-12
support for semi-colon delimited house numbers

missinglink:
fixes: https://github.com/pelias/openstreetmap/issues/21

note: I had to refactor the property mapping functions due to jshint warnings

### DIFF
--- a/stream/address_extractor.js
+++ b/stream/address_extractor.js
@@ -52,7 +52,7 @@ module.exports = function(){
 
       // accept semi-colon delimited house numbers
       // ref: https://github.com/pelias/openstreetmap/issues/21
-      var streetnumbers = doc.address.number.split(';');
+      var streetnumbers = doc.address.number.split(';').map(Function.prototype.call, String.prototype.trim);
       streetnumbers.forEach( function( streetno, i ){
 
         try {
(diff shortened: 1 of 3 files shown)


missinglink:
added support for comma delimiter as it's actually suggested here: http://taginfo.openstreetmap.org.uk/keys/addr:housenumber#values although the default delimiter for values is semi-colon, so we support both.

stream/address_extractor.js
```diff
@@ -50,42 +50,38 @@ module.exports = function(){
...
-
-        [ 'name', 'number', 'street', 'zip' ].forEach( function ( prop ){
-          try {
-            record.setAddress( prop, doc.getAddress( prop ) );
-          } catch ( ex ) {}
-        });
+      // accept semi-colon and comma delimited house numbers
+      // ref: https://github.com/pelias/openstreetmap/issues/21
+      var streetnumbers = doc.address.number.split(/[;,]/);
+      for( var i in streetnumbers ){
```
sevko: Can you use a `.forEach()` here, or at least a conventional for loop (`for(var i = 0; // etc`)?

missinglink:
```bash
$ npm publish
+ pelias-openstreetmap@2.1.0
```

PR #66, opened 2015-06-26
Get center fix

missinglink:
This PR is to fix our code which was broken by an API change to the `geolib` library we depend on:
https://github.com/manuelbieh/Geolib/commit/3b55f9749cb7d4fd2320213d0437fcf19e8e9680

The error that we were getting in the logs was:
```bash
error: [openstreetmap] failed to denormalize way Error: invalid document type, expecting: number, got NaN: NaN
```

I've updated the dependency versions, in the process `precommit-hook` was updated and added some auto-generated bits in `package.json` that were not by my hand.

The main change is in `util/geoJsonCenter.js` which now uses a spherical projection to compute the weighted centroid of the polygon, where in the past we simply computed the bbox of the polygon (from it's extremes) and then averaged the values.

The result is a better!? centroid calc which uses more CPU but accounts for the curvature of the earth, see the diff for: `test/fixtures/vancouver.extract.expected.json` for more detail.

I'm hoping to get rid of this calc completely with this https://github.com/pelias/pbf2json/issues/22 but for now this will get OSM building again.

Bonus: I managed to delete `util/geoJsonTypeFor.js` which was not unit tested and scheduled for deletion anyway.


dianashk:
:+1:

missinglink:
```bash
$ npm publish
+ pelias-openstreetmap@2.5.0
```

PR #73, opened 2015-07-06
migrate suggester from FST to ngrams

missinglink:
migrate suggester from FST to ngrams


missinglink:
```bash
$ npm publish
+ pelias-openstreetmap@3.0.0
```

PR #76, opened 2015-07-17
upgrade pbf2json

missinglink:
bring osm importer up-to-date with latest changes to `osm2pbf`, see https://github.com/pelias/pbf2json/pull/23

the gist of the changes is that we use the precomputed centroid passed in from the Go library where available to avoid doing those calcs in the nodejs process.

there is also some cleaning up and adding of tests to cover the centroid functionality

the result should be a faster and more stable build, plus we can potentially remove the one last piece of code in this repo which is not very well tested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "is-object": "^1.0.1",
     "iso-639-3": "^0.2.0",
     "merge": "^1.2.0",
-    "pbf2json": "^2.4.0",
+    "pbf2json": "^3.0.0",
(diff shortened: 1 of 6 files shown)


hkrishna:
:+1:

missinglink:
```bash
$ npm publish
npm http request PUT https://registry.npmjs.org/pelias-openstreetmap
+ pelias-openstreetmap@3.1.0
```

PR #77, opened 2015-07-20
adding burrito

hkrishna:
because who doesn't love burrito, am I right @missinglink ?


missinglink:
I'm guessing only the gluten-intolerant and they can just unwrap it, am I right? :)

This list was originally sourced from http://wiki.openstreetmap.org/wiki/Key:cuisine so you should probably update the wiki too, so mappers [start using it](http://taginfo.openstreetmap.org/search?q=cuisine%3Aburrito).

:+1:

burritojustice:
approved.

unwrapping: not approved

PR #82, opened 2015-10-14
Use Address-deduplicator

hannesj:
Hi,

Currently there is no possibility of using the address-deduplicator on openstreetmap addresses, as you can do with openaddresses. This adds the possibility of using pelias/address-deduplicator in the input stream to filter out already-seen addresses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ You can optionally change:
   county, etc. names), but you can optionally create it via the
   [`pelias/admin-lookup`](https://github.com/pelias/admin-lookup) module; just set this property to `true`.  Consult
   the `admin-lookup` README for setup documentation (namely just downloading the Quattroshapes dataset).
+- `import.openstreetmap.deduplicate` - this makes it possible to import multiple datasets containing the same addresses, without getting duplicates in the pelias database.
+  Please see [`pelias/address-deduplicator`](https://github.com/pelias/address-deduplicator) for more details.
 
 If your paths point to an SSD rather than a HDD then you will get a significant speed boost, although this is not required.
 
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@
 **/
 
 var elasticsearch = require('pelias-dbclient'),
+    addressDedupStream = require( 'pelias-address-deduplicator' );
     adminLookup = require('pelias-admin-lookup'),
     dbmapper = require('./stream/dbmapper'),
     peliasConfig = require( 'pelias-config' ).generate();
@@ -30,8 +31,13 @@ osm.import = function(opts){
     pipeline = pipeline.pipe( adminLookup.stream() );
   }
 
+  pipeline = pipeline.pipe( osm.address.extractor() );
+  
+  if( peliasConfig.imports.openstreetmap.deduplicate ){
+    pipeline = pipeline.pipe( addressDedupStream() );
+  }
+  
   pipeline
-    .pipe( osm.address.extractor() )
     .pipe( osm.category.mapper( osm.category.defaults ) )
     .pipe( dbmapper() )
     .pipe( elasticsearch() );
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "iso-639-3": "^0.2.0",
     "merge": "^1.2.0",
     "pbf2json": "^3.0.0",
+    "pelias-address-deduplicator": "^1.x.x",
     "pelias-admin-lookup": "^2.0.8",
     "pelias-config": "latest",
     "pelias-dbclient": "0.0.9",


dianashk:
@hannesj, thank you so much for the PR. This looks great. We will deploy it to our dev environment next time we reindex the planet and make sure all goes smoothly. If so, we'll merge it.

orangejulius:
Hey @hannesj,
Thanks for adding this, it's a great additional option to have, and the config flag was a nice touch.

I have a question though: how often do you see dupes in OSM data? I took a look at our Boston metroextract as an example. Without deduplication there were 14626 items imported, with deduplication there were 12091.

However, a lot of the nodes and ways removed by the deduplicator are not real duplicates. It turns out the address deduplicator defaults to a [geohash precision of 5](https://github.com/openvenues/address_deduper/blob/master/address_deduper/config.py#L52), which ends up corresponding to about a 15km square due to the way the deduper works.

This is pretty big so in Boston for example, a lot of subway stops and chain stores (especially Dunkin' Donuts, which is very popular in Boston) are removed.

Are there other places where OSM data isn't so good and a deduplicator is needed? Or are you telling the address deduplicator to use a higher precision? Very curious as it would help us decide how/if to roll this out in our Pelias instances.

--------
I did a bunch of analysis for this actually, so for you or anyone else I'll post my method here.

1. Import the data with or without the deduplicator config settting (and a previously empty Elasticsearch index)
2. dump all the data as JSON with this query (probably will break if run against a much larger dataset)

    ```bash
    curl -XPOST 'localhost:9200/pelias/_search' -d '
    {
      "query": {
         "match_all": {}
       },
       "size": 20000,
       "sort": ["name.default"]
    }' > duplicated-es.json # change this to be correct for which setting you used
    ```

3. repeat for the other config setting
4. parse the json into an array with only the name and lat/lon of each object using a script like this:

    ```javascript
    var fs = require('fs');
    var file = fs.readFileSync(process.argv[2]);
    var json = JSON.parse(file);

    var names = json.hits.hits.map(function(hit) {
      hit = hit['_source'];
      return hit.name.default + " " + hit.center_point.lat + "," + hit.center_point.lon;
    }).sort();

    console.log(names);
    ```
5. diff the two outputs of that script and you'll get something like this

    ```diff
    --- duplicated.txt      2015-10-29 15:12:21.764554137 -0400
    +++ deduplicated.txt    2015-10-29 15:14:55.880314347 -0400
    @@ -1,5 +1,4 @@
     [ '0 Charles Street 42.354356,-71.068275',
    -  '0 New Washington Street 42.380042,-71.082336',
       '1 42.358237,-71.092337',
       '1 Adamson Street 42.358451,-71.136925',
       '1 Akron Street 42.325887,-71.086411',
    @@ -7,15 +6,10 @@
       '1 Austin Park 42.367092,-71.104063',
       '1 Avenue De Lafayette 42.353404,-71.060771',
       '1 Batterymarch Park 42.229922,-71.02562',
    -  '1 Batterymarch Park 42.229922,-71.02562',
       '1 Beacon Street 42.358394,-71.060756',
       '1 Belgrade Avenue 42.287487,-71.130317',
       '1 Belmont Street 42.374856,-71.150301',
    -  '1 Broadway 42.362634,-71.083658',
    -  '1 Broadway 42.362665,-71.083477',
       '1 Broadway 42.362744,-71.083786',
    -  '1 Broadway 42.362771,-71.083278',
    ```

Here's the full file result
[differences.txt](https://github.com/pelias/openstreetmap/files/23719/differences.txt)

hannesj:
Hi @orangejulius,

We are mainly using the deduplicator to remove addresses from OSM that are already imported from other sources, mostly OpenAdresses. We run both of the imports against the same deduper instance, so addresses that have already been imported from OpenAddresses do not get duplicated into our database. It would of course be possible to exclude those during runtime, but our preference has been to put only the most accurate data into the database.

Thanks for the hint about the geohash accuracy, we definitely would like to tune down the value in order to not miss any addresses in neighbouring municipalities.

riordan:
That's a really interesting approach. We'll see how we can incorporate
parts of it into ours.

On Thu, Oct 29, 2015 at 4:08 PM, Hannes Junnila <notifications@github.com>
wrote:

> Hi @orangejulius <https://github.com/orangejulius>,
>
> We are mainly using the deduplicator to remove addresses from OSM that are
> already imported from other sources, mostly OpenAdresses. We run both of
> the imports against the same deduper instance, so addresses that have
> already been imported from OpenAddresses do not get duplicated into our
> database. It would of course be possible to exclude those during runtime,
> but our preference has been to put only the most accurate data into the
> database.
>
> Thanks for the hint about the geohash accuracy, we definitely would like
> to tune down the value in order to not miss any addresses in neighbouring
> municipalities.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pelias/openstreetmap/pull/82#issuecomment-152303960>.
>



-- 
David Riordan | Product Manager - Search | dave.riordan@mapzen.com |
@riordan <https://twitter.com/riordan> | gpg 235D9DC95EF6277C
<https://keybase.io/riordan>
Mapzen | https://mapzen.com | @mapzen <https://twitter.com/search>

hannesj:
Also, some duplication exists in OSM, if you check eg. "Aapelinkatu 5" around [this place](http://pelias.github.io/leaflet-geocoder/#17.35000000000009/60.15543/24.75024)

PR #89, opened 2016-01-19
Switch to new model and WOF admin lookup

dianashk:
- updated Document constuctor usage to support new interface
- updated integration end-to-end tests (see [gist](https://gist.github.com/dianashk/1f437bf738f09c38769e) for fixture conversion script)
- refactored index.js to be cleaner and more approachable
- moved streams that can be enabled/disabled into a separate module to hide complexity, this also means there are now empty passthrough streams in place of disabled streams

Fixes pelias/pelias#227


stream/deduper.js
```diff
@@ -0,0 +1,18 @@
...
+var addressDedupeStream = require( 'pelias-address-deduplicator' );
+
+
+function createStream() {
+  // if deduping is enabled, create the deduper stream
+  // otherwise create a passthrough stream that does nothing
+  if (peliasConfig.imports.openstreetmap.deduplicate) {
+    return addressDedupeStream();
+  }
+  else {
```
orangejulius: This is cool. The pass-through streams makes the main code that assembles the pipes much cleaner.

orangejulius:
:+1:

PR #90, opened 2016-01-25
Update id building scheme and add logging

dianashk:
New ids should be as follows:

for named POI (venue): 
  `_id` = `[node | way]:[osm-id]`
  `_type` = `venue`
  `data.category` = will have categories

for addresses: 
  `_id` = `[node | way]:[osm-id]`
  `_type` = `address`
  `data.category` = will have __NO__ categories

for addresses with multiple house numbers: 
  `_id` = `[node | way]:[osm-id]:[housenum]`
  `_type` = `address`
  `data.category` = will have __NO__ categories

Fixes pelias/pelias#227

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,10 @@ As a minimum, it should be valid json and contain the following:
   },
   "imports": {
     "openstreetmap": {
-      "adminLookup": true,
+      "adminLookup": false,
       "leveldbpath": "/tmp",
       "datapath": "/data/pbf",
       "import": [{
-        "type": { "node": "osmnode", "way": "osmway" },
         "filename": "london_england.osm.pbf"
       }]
     }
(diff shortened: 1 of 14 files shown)


orangejulius:
:+1: 

Minor idenote, it looks like there are some tests and files that are consistent, but using old values for source/layer/id etc. A bit of a pain but they might be nice to fix to avoid confusion. Try these to find them:

`git grep osm`
`git grep osmway`
`git grep osmnode`

Examples:
* https://github.com/pelias/openstreetmap/blob/fix-ids/test/stream/category_mapper.js#L43
* https://github.com/pelias/openstreetmap/blame/fix-ids/README.md#L100

Also, question:
Is the type object [shown here](https://github.com/pelias/openstreetmap/blame/fix-ids/README.md#L53) still used? I can't figure out where it feeds into the code.

trescube:
:+1:

PR #103, opened 2016-02-22
Add bbox computation to ways.

dianashk:
Adding bounding box computation to `way` type records from OSM. Records of type `node` will remain unchanged.

Fixes pelias/pelias#243


trescube:
:+1:

PR #109, opened 2016-03-02
Update pelias-wof-admin-lookup and remove process.exit temp hack

dianashk:
Connected to pelias/wof-admin-lookup#21


orangejulius:
Hooray! :+1:

PR #135, opened 2016-04-21
Increment package version to 4.0.0

dianashk:


dianashk:
```
$ npm publish
+ pelias-openstreetmap@4.0.0
```

orangejulius:
PRs with version increments are kinda weird because what if I thumbs down this? you can't undo the publish :(

orangejulius:
but, also :+1:

dianashk:
yeah, I realized that after seeing the greenkeeper PR that was failing as soon as I moved this into the review column. Sorry. Let's discuss the right way to do the version increments.

PR #141, opened 2016-06-10
Check the proper admin lookup config properties

orangejulius:
The admin lookup stream code was incorrectly checking the wrong
property, making it harder to enable admin lookup.

Thanks to @DylanFrese for helping us figure this out.

Related: pelias/wof-admin-lookup#52
Fixes pelias/wof-admin-lookup#51


trescube:
👍

PR #149, opened 2016-07-06
Update travis

orangejulius:
The usual updates that we have been slowly rolling out to other repos:

* Add Node 6, drop Node 0.10
* Only specify major versions for Node, so the latest minor release gets used automatically
* Cache node_modules dir to speed things up
* Add a travis NPM script that runs the unit tests and end-to-end tests. this speeds things up by reducing the number of times a new Travis worker has to start up
* Reorder all the NPM scripts in the package.json file alphabetically

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,30 @@
 sudo: false
 language: node_js
+cache:
+  directories:
+    - node_modules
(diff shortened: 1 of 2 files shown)


trescube:
👍

PR #152, opened 2016-07-08
Cleaning up code for OSM importer

avulfson17:
This is mostly just a stylistic change. The code makes more sense this way
(for me at least)


orangejulius:
Definitely a good change. Reduces duplication and makes the code read a bit better. LGTM

dianashk:
:shipit:

PR #156, opened 2016-07-19
Refactor streams interface

### LINKED ISSUE #55: Import multiple pbf files
originally reported by @loomsen in https://github.com/pelias/openstreetmap/issues/46

currently the `pelias/config` structure implies that a consumer can specify multiple `pbf` files to be imported; however the import script does not currently support this.

---

Thanks, all seems to be working fine now. However, I wonder if it's possible to import multiple pbf files?
I've tried adding another one into the json array file, but it doesn't seem to be imported. Only the first entry seems to be processed. If I import it afterwards, data from both show up, so this would work. Would be nicer to configure an array though.

```json
      "import": [{
        "type": { "node": "osmnode", "way": "osmway" },
        "filename": "germany-latest.osm.pbf"
      },
      {  
        "type": { "node": "osmnode", "way": "osmway" },
        "filename": "liechtenstein-latest.osm.pbf"
      }]
```

avulfson17:
This refactors the importer code, including changing the index file to basically only run the importer, and allowing the pbfparser to be called with no arguments.

Not 100% sure that all the branch/tree history (or whatever its called) is correct, but itd be nice if i could get confirmation that it is

Fixes #55


dianashk:
Let's fix the travis tests and also remove the excessive logging that's happening during the testing.

dianashk:
Fixes #55

package.json
```diff
@@ -58,12 +58,14 @@
   "devDependencies": {
     "async": "^2.0.0",
     "colors": "^1.1.2",
+    "combined-stream": "^1.0.5",
```
dianashk: I think `combined-stream` should be under regular `dependencies`, not `devDependencies`. `devDependencies` don't get installed in a production environment, so this could lead to big trouble.

stream/multiple_pbfs.js
```diff
@@ -0,0 +1,20 @@
...
+var path = require('path');
+
+function createCombinedStream(){
+  var fullStream = combinedStream.create();
+  var defaultPath= require('pelias-config').generate().imports.openstreetmap;
+
+  defaultPath.import.forEach(function( importObject){
+    var conf = {file: path.join(defaultPath.datapath, importObject.filename), leveldb: defaultPath.leveldbpath};
+    fullStream.append(function(next){
+      console.log('Creating read stream for: ' + conf.file);
```
dianashk: use `pelias-logger` instead of `console.log` (check other places where it might need to be replaced)

dianashk:
This looks really great. The refactoring makes me so happy! Just a few minor comments and it will be ready for primetime.

avulfson17:
I dont really know when/why to use the logger but the 2 cases i found console.log used were here (this is the end-to-end test failures output):
`if( diff ){
   97        //added for clarification because apparently understanding the deep diff output is hard
   98:       console.log('actual:', JSON.stringify(actual[0], null, 2));
   99:       console.log('expected:', JSON.stringify(expected[0], null, 2));
  100  
  101:       console.log( JSON.stringify(diff, null, 2) );
  102:       console.log('actual count:', actual.length);
  103:       console.log('expected count:', expected.length);
  104:       console.log('matching count:', colors.green(countSame));
  105:       console.log('difference count:', colors.red(countDiff));
  106        console.error( colors.red( 'end-to-end tests failed :(' ) );
  107        console.error( 'contents of', tmpfile, 'do not match expected:', expectedPath );`

and here (this looks like its supposed to print out the tags):
`50    // print out
   51:   console.log( JSON.stringify( tags, null, 2 ) );
   52  });
`

dianashk:
No need to update test code with pelias-logger, just stick to the production code. There may not be a whole lot of places, but it's worth taking a look. I think just the one place I called it out in should be fine. Thanks.

orangejulius:
It looks like this pull request changed at some point from a refactoring (strict definition: making code do exactly the same thing, but in a better/cleaner way) to adding a new feature. Which is of course great, but it seems like it might make sense to create a new pull request with a branch name and pull request name that describe adding the new feature

avulfson17:
Is there a way to move some of the commits into a new branch or something like that? I know exactly when this break happens but I dont really know the best way to deal with it. I was just going to close #55 with a comment linking to this pull request.

orangejulius:
In this case it's probably more hassle than it's worth to split the two changes. Performing initial refactoring to make it easier to add a new feature is A-OK, and quite common.

But creating a new branch and PR is helpful to keep track of what actual changes happened, and should definitely be done here.

avulfson17:
I dont really understand what you want me to do here. Do you want me to create a dummy branch that all these changes get pushed to and then push it to master? Do you want me to recreate all the changes in a totally new PR? Or something else entirely?

orangejulius:
Replaced by #163

PR #172, opened 2016-08-15
Upgrade to pbf2json 4.1.0

orangejulius:
Also require an exact version of pbf2json rather than a caret range.

This is in line with https://github.com/pelias/pelias/issues/366, where
we decided to only use exact version deps for other Pelias code. This
both helps us track changes better, and ensure we can know exactly what
code we're releasing at any time.

Connects pelias/pelias#366


avulfson17:
LGTM

PR #173, opened 2016-08-16
Require exact version of pbf2json

orangejulius:
This is in line with https://github.com/pelias/pelias/issues/366, where
we decided to only use exact version deps for other Pelias code. This
both helps us track changes better, and ensure we can know exactly what
code we're releasing at any time.


missinglink:
:shipit:

trescube:
:shipit:

PR #221, opened 2017-01-03
Add config validation

trescube:
Required fields:

* `datapath`
* `import` (the files to import, can be empty)
* `leveldbpath`

Optional fields:

* `adminLookup`
* `deduplicate`

If any of the optional are actually required, let me know. 

Also bumped lint to support es6.


orangejulius:
`leveldbpath`, `adminLookup` and `deduplicate` are all required I think, although they have default values. `adminLookup` and `deduplicate` are booleans that default to false.

trescube:
made `leveldbpath` required.  if `adminLookup` and `deduplicate` have defaults then they're not required.

orangejulius:
I guess that makes sense. It's very hard for a boolean value that's missing to cause problems.

PR #236, opened 2017-02-09
removed adminLookup in favor of wof-admin-lookup making the decision

trescube:


trescube:
makes it easy to review!

PR #280, opened 2017-05-22
feature whitelist - re: aeroway

missinglink:
reduce target `aeroway` values to only those enumerated in a whitelist

see issue https://github.com/pelias/pelias/issues/573 for more detail.

this PR changes our broad `aeroway+name` filter to a more selective one.

more notes on the issue linked above.

closes https://github.com/pelias/pelias/issues/573

### DIFF
--- a/config/features.js
+++ b/config/features.js
@@ -25,7 +25,6 @@ var tags = [
   'craft+name',
   'military+name',
   'aeroway~terminal+name',
-  'aeroway~gate+name',
   'aeroway~aerodrome+name',
   'aeroway~helipad+name',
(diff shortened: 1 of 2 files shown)


missinglink:
I'm now thinking we drop `'aeroway~gate+name'`, the gates names are noisy, generally things like `M14` or `A`, `B` etc.

Since they're only really applicable for indoor routing in an area that is generally navigated by signage, I'm thinking they would more often be a hinderance than a help.

missinglink:
added two commits:

- remove 'aeroway~gate+name'
- improved tests

PR #286, opened 2017-06-13
Increase Node.js memory limit

orangejulius:
The admin lookup worker processes are starting to hit the default memory
limit. Increasing it gives us time to improve the admin lookup
architecture.

Connected to https://github.com/pelias/wof-admin-lookup/issues/117

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "end-to-end": "npm run pretest && node test/end-to-end.js;",
     "lint": "jshint .",
     "pretest": "test/pretest.sh",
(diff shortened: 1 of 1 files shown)


missinglink:
@orangejulius did you also intent to change `index.js` to `import.js` or was that a typo?

orangejulius:
@missinglink oh definitely not, good catch. I used the same commit from openaddresses to keep the commit message similar, but I forgot we hadn't standardized on the entrypoint file name. Maybe someday we will. Fixed now!

PR #294, opened 2017-07-07
Add download script

dianashk:
Connected to pelias/pelias#506

Add download script that will download any files listed in the new config section.
Config can now contain `imports.openstreetmap.download` which is an array of objects with `sourceURL` properties.
If the download config option is not specified, the entire latest planet file will be downloaded.
All downloaded files are saved to the path specified by `imports.openstreetmap.datapath`.


util/download_data.js
```diff
@@ -0,0 +1,57 @@
...
+      process.exit(1);
+    }
+    logger.info('All done!');
+  });
+}
+
+function download(callback) {
+  let sources;
+
+  if (!_.isEmpty(config.imports.openstreetmap.download)) {
```
orangejulius: if you swap the two clauses of this if/else you can get rid of the negation operator and the code reads slightly cleaner. plus we need to save those exclamation points for when we are excited about something! :P

util/download_data.js
```diff
@@ -0,0 +1,57 @@
...
+    sources = _.map(config.imports.openstreetmap.download, (source) => source.sourceURL);
+  }
+  else {
+    sources = [
+      'http://planet.openstreetmap.org/pbf/planet-latest.osm.pbf'
```
orangejulius: So on the one hand, the planet file makes sense as a default because it's what we would use in production, and because it has everything. But it's also really really really big. I could imagine someone just going with the defaults and being very confused when their local machine crumbles under the weight of trying to import all of OSM.

I'm not sure there's a better solution though. We could link to one of our smaller "standard" extracts like Berlin, London, NYC, etc, but then we are defaulting to data that isn't relevant to everyone.

There might not be an ideal solution, anyone have any ideas?

util/download_data.js
```diff
@@ -0,0 +1,57 @@
...
+      callback
+    );
+  });
+}
+
+function downloadSource(config, sourceUrl, callback) {
+  const targetDir = config.imports.openstreetmap.datapath;
+
+  logger.debug(`downloading ${sourceUrl} to ${targetDir}`);
+  child_process.exec(`cd ${targetDir} && { curl -L -X GET --silent --remote-name ${sourceUrl}; cd -; }`, callback);
```
orangejulius: What does the `cd` at the end accomplish? Shouldn't each child_process exec inherit the working directory from wherever the importer is started, and thus only require the first `cd`?

README.md
```diff
@@ -28,6 +28,14 @@ $ npm install
 ## Download data
```
tigerlily-he: Now there are two ways of downloading data: (1) manually from Metro Extracts, OSM, or Geofabrik or (2) via specifying URL in config and running `npm run download`. 

For the first option, the `openstreetmap.download` property is irrelevant. And you don't need to run `npm run download`.

For the latter method, it makes more sense to see the "Configuration" section before running `$ PELIAS_CONFIG=<path-to-config> npm run download`. `imports.openstreetmap.datapath` specifies the directory which the pbf file _will be_ downloaded to.

I think the instructions would be clearer if "Download Data" split into the "Manually" and "Using Download Script" and each included the specific steps and example config file. Or, the sections can remain as is but with notes on which step is necessary for which option.

schema.js
```diff
@@ -15,6 +16,9 @@ module.exports = Joi.object().keys({
       import: Joi.array().items(Joi.object().keys({
         filename: Joi.string()
       }).requiredKeys('filename').unknown(true)),
+      download: Joi.array().items(Joi.object().keys({
+        sourceURL: Joi.string()
```
tigerlily-he: If the source URLs are all `http` or `https`, then the Joi schema can be stricter: `Joi.string().uri({ scheme: /https?/ })`

orangejulius: A lot of geo people are old-school. someone's going to want `ftp` :P

util/download_data.js
tigerlily-he: Default to nothing and throw error that a source needs to be specified?

tigerlily-he:
Downloads are much more convenient now because everything is centralized in the config! It doesn't overwrite existing files, which is great. 

The download data script works as long as the sourceURL is valid.  I misspelled a "london" in the URL (https://s3.amazonaws.com/metro-extracts.mapzen.com/lndon_england.osm.pbf) and `lndon_england.osm.pbf` got downloaded (it just contains html error).

Also, I tried throwing `https://mapzen.com/` into the `sourceURL` field and still got the download success message "all done". If no file got downloaded, an error message can be thrown.

util/download_data.js
dianashk: Yeah, I struggled with this a bit too, but ultimately the entire planet IS the default. :) Hopefully, people can rely on the docs to keep them out of these potential pitfalls. Also, this can always be updated in the future if we get feedback from the community.

schema.js
dianashk: good idea, @tigerlily-he, but unfortunately I think @orangejulius is right about some folks still relying on older tools.

dianashk:
### Update:

Thanks all for the thoughtful feedback. I believe I've addressed all of it with the latest commits. Please have another look.

dianashk:
Oops, missed the README suggestions from @tigerlily-he. Will take care of it now.

dianashk:
Merging and opening a separate issue to clean up download documentation.

PR #296, opened 2017-07-07
fix(package): update pelias-model to version 4.10.0

orangejulius:
Closes #293


orangejulius:
@trescube I cherry-picked your PR to fix the fixtures and put it in this branch. this one should pass now and we can avoid having to merge two PRs for pelias-model per repo

PR #306, opened 2017-07-18
fix(package): update pelias-dbclient to version 2.1.2

orangejulius:


orangejulius:
Merging without review so no one has to bother. Feel free to come in later with questions

PR #339, opened 2017-10-03
Put verbose point logging behind a config flag

orangejulius:
This importer currently logs _every_ record imported as well as its name and centroid. While useful for debugging, it's not needed all the time, even at the `verbose` logger level.

By putting this option behind a `pelias.json` config flag (defaulting to off) we can have it when we need it, and save over 10GB in each of our build logs when we don't.


missinglink:
I would be :+1: with removing the line completely and :-1: introducing a config option which is confusingly similar to the existing log level option.

orangejulius:
Okay great. I actually would prefer to remove it too, but wasn't sure if others wanted the ability to turn it back on. It's already on the verbose log level, and is still far too noisy even for that. I'll make a new PR.

PR #346, opened 2017-10-11
Move `async` to dependencies from devDependencies

### LINKED ISSUE #312: async listed as a 'dev dependency'
async is listed as a 'dev dependency', so trying to run `npm run download` after `npm install --production` results in `Error: Cannot find module 'async'`.

it might be better to move this (and any other associated dependencies) in to the main dependency list if they are intended to be used in production.

tadjik1:
Fix #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "elasticsearch": ">=1.3.4"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


orangejulius:
Nice, thanks for picking that up

PR #362, opened 2017-10-23
Upgrade to pelias-wof-admin-lookup 4.3.0

orangejulius:
This is required for HTTP PIP support!


orangejulius:
Self-merging so we can kick off an HTTP PIP dev build.

PR #378, opened 2017-11-21
filter railway features

### LINKED ISSUE #377: Remove a few railway features from the import
Hi,

With [this request](https://mapzen.com/search/explorer/?query=search&text=gare%20de%20lyon), I got 2 results that are rail tracks (`railway = rail`) :
* https://mapzen.com/search/explorer/?query=place&ids=openstreetmap%3Avenue%3Away%3A332205985
* https://mapzen.com/search/explorer/?query=place&ids=openstreetmap%3Avenue%3Away%3A115843882

I think it does not really make sense to get the named rail tracks from the OSM import.
Is it ok to change the features list to be more restrictive ?

I'm thinking of replacing
 `railway+name`

by

* `railway~station+name`
* `railway~tram_stop+name`
* `railway~halt+name`
* `railway~subway_entrance+name` 
* `railway~train_station_entrance+name`

Would that be ok ?

poke @missinglink

nlehuby:
closes #377


nlehuby:
For now, I have tested it only in Paris city.

missinglink:
hey @nlehuby thanks for the PR, this looks great to me, I have one question and one issue that needs be resolved before we can merge this.

Question: You added 5 lines to the features list and only test two of them, you also chose not to test that `railway+name` is no longer present in the array, which is probably what you're trying to avoid someone adding it back in the future? :)

Blocker: The Travis tests are failing which is blocking the merge, we will need to resolve this issue, the test suite runs the command `npm run travis` which you can run locally to reproduce the failure.

It looks like the issue is a fixture file, we have two fixtures `vancouver_canada` and `queens_village_ny` which will be downloaded automatically when you run that command.

The test suite uses those files as an input and then compares the new output vs. the results we expected from the last run.

I imagine that the expected file now contains some `railway` lines which now need to be removed.

I remember the diff checking that's part of the test suite was quite verbose and a little annoying, if you get stuck then let me know and I can help out.

missinglink:
I just had a look at the diff thing and *it is* verbose and annoying, the easiest way to handle this is:

```
npm run travis # this will fail
ls *_*.json # and produce these two files (which you can later delete)
cp actual_output.json test/fixtures/combined_vancouver_queens.json
git diff
```

this should show you the changes you have made to the Vancouver and Queens data, I had a quick look and it was what you were expecting.

if you are happy with the result then please commit the updated `test/fixtures/combined_vancouver_queens.json` file and push it to your branch.

if all is well the Travis tests will pass.

missinglink:
I think this will be a great improvement to the code, it will remove a lot of noise from the database.

nlehuby:
Hi,

I did not change the tests too much because I did not understand the test workflow, and what was tested by "npm test"...
This is clearer now, I will check this out, thanks !

nlehuby:
I think this is ok now.

nlehuby:
Thank you for your mentoring :+1:

missinglink:
Regarding mapzen search, this feature will go through a series of testing in dev, staging and finally (pending any issues) will be promoted to master.

I expect the merge to go through without issue, if we have acceptance tests which expect any of these features I will handle updating them.

The process can take anywhere from 1-3 weeks depending on what else is being shipped and tested at the same time.

Anyone running their own instances will be able to rebuild now from `master` to get these changes.

PR #407, opened 2018-04-12
Remove single-file import notice from README

kylebarron:


kylebarron:
See https://github.com/pelias/openstreetmap/issues/55#issuecomment-380959487

PR #435, opened 2018-06-26
support for mapping tags to name.default aliases

missinglink:
this PR allows OSM tags to be mapped as aliases of the `name.default` field.

the `schema/name_osm.js` file contains a mapping of OSM tags to `name.*` fields.

prior to this PR we were only able to map 1:1 from OSM to the Pelias model.

this PR now supports multiple values in that object set as `default`, in this case the additional tags will be mapped as aliases of the `name.default` field.

the configuration also adds an additional inenumerable property called `_primary` which defines which of those is considered the 'primary' name.

an example tag mapping:

```
osm_tags = { name: "foo", short_name: "bar", reg_name: "baz" }

# prior to the PR
name = {
  default: "foo",
  short: "bar",
  regional: "baz"
}

# after the PR
name = {
  default: [ "foo", "bar" ],
  regional: "baz"
}
```

---

I enabled 3 tags as aliases of default: `[ short_name, alt_name, loc_name ]`, of these only `short_name` is supported by nominatim.

https://wiki.openstreetmap.org/wiki/Names#Key_Variations

The [test/fixtures/combined_vancouver_queens.json](https://github.com/pelias/openstreetmap/pull/435/files#diff-8dc60323dfa03395125c0936781cf634) file in this PR includes a diff which shows and example of the expected changes.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,6 +17,6 @@
   "undef": true,
   "unused": false,
   "maxparams": 4,
-  "maxdepth": 4,
+  "maxdepth": 5,
   "maxlen": 140
 }
--- a/schema/name_osm.js
+++ b/schema/name_osm.js
@@ -7,21 +7,43 @@
 
   eg. tags['alt_name'] -> doc.name['alternative']
 
+  The value 'default' is special, it defines the name used by default
+  for elasticsearch queries, it has the unique ability to specify multiple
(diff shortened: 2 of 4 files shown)


orangejulius:
The most recent Nextzen full planet build includes these changes, but I haven't looked at the results yet. I'm spinning up a cluster to do that now.

PR #448, opened 2018-09-04
fix(logger): remove logger line

missinglink:
remove log line missed in: https://github.com/pelias/openstreetmap/pull/347


orangejulius:
Good catch :)

PR #449, opened 2018-09-05
feat(blacklist-stream): allow excluding documents via GID blacklist

missinglink:
This PR included a new stream called [blacklist-stream](https://github.com/pelias/blacklist-stream) which allows records to be excluded from the build.

See https://github.com/pelias/blacklist-stream for more information on how to use the blacklist file.

Note: `blacklistStream` MUST come after `addressExtractor` as the `addressExtractor` can change the document `layer`.


orangejulius:
Works great!

PR #454, opened 2018-09-13
Support read-only root filesystem

orangejulius:
This PR includes changes to better support running our Docker containers with a read-only root fs.

The changes are:
- Create start and download scripts so that calls to `npm` within containers can be avoided
- Include the `vancouver_canada.osm.pbf` test file, rather than downloading it.
- Run the unit tests as the `pelias` user. This is important because LevelDB, via pbf2json, leaves some files behind in `/tmp`. It's better if those files are owned by the `pelias` user. They should probably be removed all together somehow.

Connects https://github.com/pelias/pelias/issues/745


orangejulius:
Note: if anyone was seeing errors like the following, this should have fixed it:

```
[pbf2json]: 2018/09/13 18:01:15 open /tmp/LOCK: permission denied
```

PR #464, opened 2018-11-01
upgrade pbf2json, remove code which depended on noderefs

missinglink:
lots of code deletions :)

This PR upgrades to use [version 5 of pbf2json](https://github.com/pelias/pbf2json/pull/69) which computes bounding boxes for us.

Since `pbf2json` now provides both `centroid` values and `bounds`, we no longer need it to produce the `nodes` array.
We'll likely get a performance boost by not having to serialize and deserialize this large array between programs.

Another bonus is that I have deleted all the `geolib` code which computes centroids and bounding boxes.


missinglink:
one less dependency and one less stream FTW

missinglink:
it's amazing that the `end-to-end` tests 'just work', I thought that there may have been a difference in precision, but that's not the case, even though `pbf2json` is set to truncate floats at 7 decimal places.

PR #474, opened 2019-03-21
chore(tests): fix tests, add relations to fixture file

missinglink:
fix tests which broke in https://github.com/pelias/openstreetmap/pull/471


missinglink:
circle-ci job passed here: https://circleci.com/gh/pelias/openstreetmap/508

PR #480, opened 2019-04-25
pbf2json bugfix pbf2json/pull/77

missinglink:
Updated `pelias/pbf2json` dependency to bring in https://github.com/pelias/pbf2json/pull/77

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ esclient.log
 nohup.out
 test/*.pbf
 coverage
+
+actual_output.json
+expected_output.json
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "joi": "^14.0.0",
     "lodash": "^4.17.4",
     "merge": "^1.2.0",
-    "pbf2json": "^6.0.0",
+    "pbf2json": "^6.1.0",
     "pelias-blacklist-stream": "^1.0.0",
     "pelias-config": "^3.3.0",
     "pelias-dbclient": "^2.8.0",


missinglink:
I think we might need to update the `features.js` list at some point to reduce the number of records with tags such as `natural:lake`, but that's a topic for another PR ;)

missinglink:
cc/ @Joxit

This PR fixes a bug which prevented a large amount of `relations` from being imported.

We're a little bit busy at the moment, if you get a chance to test it out we'd be happy for a review :)

Joxit:
Okay @missinglink, I'm doing a planet import with this PR :)

Joxit:
Import done, here are my stats :
138801 messages `denormalize failed for relation: xxx no ways found`; This makes sense because in these relations, there were only nodes and no ways such as [3727026](https://www.openstreetmap.org/relation/3727026)
89 messages `denormalize failed for relation: xxx no valid bounds`; these ones are a bit strange relations, I don't know how this should be processed. Example: [9119261](https://www.openstreetmap.org/relation/9119261)

101360966 elements was indexed by pelias-openstreetmap
18647884 elements was indexed by pelias-polylines build by [missinglink/pbf](https://github.com/missinglink/pbf)
120008846 elements in ES with OSM source (pelias-openstreetmap + pelias-polylines). My current production server has 122658366 elements.

missinglink:
Thanks @Joxit

1. It's a bit worrying that your build seems to have 2.6M records less records?
2. Are you using `missinglink/pbf` to build polylines for the planet? I assumed that this wouldn't work due to RAM issues, we use the [Valhalla](https://github.com/pelias/polylines/wiki/Generating-polylines-from-Valhalla) method for the planet.
3. We could probably do something about the `no ways found` and `no valid bounds` errors, we could use the lat/lon of one of the nodes where no ways exist, I'll open an issue on pbf2json for that.

Joxit:
I have my old build stats (this could help a bit more)

| Import | Old build | New Build |
| ---- | ---- | ---- |
| Polylines | 25503343 (`valhalla`) | 18647884 (`missinglink/pbf`) (**-6.8M**) |
| venues from OSM | 21999463 | 23257451 (**+1.2M**) |
| addresses from OSM | 75155565 | 78103515 (**+2.9M**)|
| Total OSM | 97155028 | 101360966 (**+4.2M**) |

The difference between old build and the new one is definitely because of polylines, I will do another build with your extract that contains 25964427 elements.
In fact I didn't get any error with pbf (in logs), only 6 warnings, I thought that would be enough.

I tried some requests such as [9087463](https://osm.org/relation/9087463) and it works well :smile:

missinglink:
Oh that's good, so it sounds like this PR is fine to merge then 🎉 

Regarding the polylines extract, you could try using this one https://geocode.earth/data, it's not too old and was generated using the Valhalla extract method.

Joxit:
Yes! 😁

Thanks, I will save this link, it will be easier for me :)

PR #490, opened 2019-06-05
feat(features): Add pedestrian squares

orangejulius:
Pedestrian squares are often well known places that are essentially venues. This change allows any OSM record with the tags `highway=pedestrian`, `area=yes`, and a name to be imported as a venue.

### DIFF
--- a/config/features.js
+++ b/config/features.js
@@ -22,6 +22,8 @@ var venue_tags = [
   'railway~halt+name',
   'railway~subway_entrance+name',
   'railway~train_station_entrance+name',
+  'highway~pedestrian+area~yes+name',
+  'place~square+name',
   'sport+name',
   'natural+name',
   'tourism+name',


missinglink:
It would be great if we could include this too?
https://www.openstreetmap.org/relation/6137529

missinglink:
And this one too :)
https://www.openstreetmap.org/relation/3573535

Maybe by targeting the tourism tag?

missinglink:
https://www.openstreetmap.org/relation/6829553

orangejulius:
We already import both of those because we do look at the [tourism tag](https://github.com/pelias/openstreetmap/blob/master/config/features.js#L27).

After spending some time looking at the OSM wiki I think what we will miss without this PR is pedestrian squares that are not "major tourist attractions".

missinglink:
This one is actually tagged 'place:square'
https://www.openstreetmap.org/relation/131761

orangejulius:
Oh interesting. Well, as usual OSM has many ways to do the same thing. According to taginfo there are only 22k places tagged `place=square` but over 500k `highway=pedestrian`. I don't see any harm in supporting both though.

missinglink:
Makes sense

There is one change to the fixture file which isn't ideal, it's a 'bus loop'.

Is that something we want? Would it be possible to avoid elements like that?

orangejulius:
Good catch. I'm going to say [that place](https://www.openstreetmap.org/way/99144094) is tagged incorrectly. It's definitely a bus stop and it should not have `highway=pedestrian` at all. It won't affect our tests unless we update the included PBF file, but I'm going to fix it on OSM.

missinglink:
Makes sense, I think you're right

orangejulius:
I added support for `place=square` just to handle all the cases. Its so rare it doesn't even affect the output from the fixture PBF files.

Issue #385, opened 2018-01-05
Population info


tadjik1:
Hi there,

we've noticed that there is a sorting mechanism in `pelias-api` that takes `population` field into account. At the same time seems that this information is missing for OSM documents. Am I missing something here or this feature is in "todo" list and it should be implemented in importers (by calling placeholder service, for example)?

missinglink:
We currently only have population information for administratrative areas such as cities and states, the field doesn't really make sense for other layers.

missinglink:
The popularity field is more appropriate for venues, streets etc.

tadjik1:
@missinglink and what do you think of using `population` also for streets, pois, etc. it will be population let's say of `county` in which that place is located. so that if there will be several streets with the same name or two mcdonalds then population of district will be taken into account.

missinglink:
@tadjik1 It depends on how it's implemented, a user in a rural area should still be able to find a local chain restaurant (eg. McDonalds) even if there are hundreds in a nearby city.

Some care would need to be taken to ensure that the population calculation didn't outweigh the proximity calculation.

There are also some other tricky cases such as remote tourist attractions (eg. something like Angkor Wat, in a sparsely populated area of Cambodia), which also have replicas in densely populated cities such as Las Vegas.

It all depends on the use-case, I'm not sure if population of the encompassing geography adds to the 'value' of the results, other metrics such as `focus.point` would probably provide more meaningful results to the user.

missinglink:
I would be interested in using `popularity` for streets, this is something other geocoders do, so if you type just 's' you get back the top 5 most trafficked roads starting with 's'.

missinglink:
I'm planning to come in to the Moovel office tomorrow so we can talk more then :)

PR #493, opened 2019-06-12
venue popularity

missinglink:
This feature has been requested in https://github.com/pelias/pelias/issues/171

This PR adds the ability to increase a document `popularity` score based on which tags it has.

I'd love to see some suggestions from the community on what they would like to see regarding 'importance' scoring in OSM.
Please suggest additional tags and scoring methodologies!

I will tune the final numbers in testing, these numbers are normalized using `log1p` before the scoring is applied in elasticsearch.
We should only really use these numbers as a 'tie-breaker' for multiple venues with the same name, eg `Eiffel Tower`

I also added some functionality to detect `abandoned` and `disused` places and give them negative popularity.
In the case where a document has negative popularity, it is discarded.

resolves https://github.com/pelias/pelias/issues/171

### DIFF
--- a/stream/popularity_mapper.js
+++ b/stream/popularity_mapper.js
@@ -1,0 +1,199 @@
+/**
+  The popularity mapper is responsible for generating a 'popularity'
+  value by inspecting OSM tags.
+
+  Disused and abandoned places are given a strong negative score.
+  If the popularity score is less than zero then the document is discarded.
+
+  Feel free to make changes to this mapping file!
+**/
+
+const through = require('through2');
+const peliasLogger = require('pelias-logger').get('openstreetmap');
+
+const config = {
+  // https://taginfo.openstreetmap.org/keys/importance
+  importance: {
+    international: { _score: 50000 },
+    national: { _score: 10000 },
+    regional: { _score: 5000 },
+    urban: { _score: 1000 },
+    suburban: { _score: 500 },
+    local: { _score: 100 },
+  },
+
+  // concordances
+  wikipedia: { _score: 3000 },
+  wikidata: { _score: 3000 },
+
+  // properties found on major landmarks &
+  // public buildings.
+  architect: { _score: 5000 },
+  heritage: { _score: 5000 },
+  'heritage:operator': { _score: 2000 },
+  historic: {
+    building: { _score: 5000 },
+    church: { _score: 5000 },
+    _score: 1000,
+  },
+  building: {
+    colour: { _score: 2000 },
+    material: { _score: 2000 },
+    supermarket: { _score: 2000 },
+    civic: { _score: 2000 },
+    government: { _score: 2000 },
+    hospital: { _score: 2000 },
+    train_station: { _score: 5000 },
+    transportation: { _score: 2000 },
+    university: { _score: 2000 },
+    public: { _score: 2000 },
+    sports_hall: { _score: 2000 },
+  },
+  height: { _score: 2000 },
+  start_date: { _score: 2000 },
+  opening_date: { _score: 2000 },
+
+  // properties found on tourist attractions
+  tourism: {
+    visitors: { _score: 2000 },
+    aquarium: { _score: 2000 },
+    attraction: { _score: 1000 },
+    museum: { _score: 2000 },
+    theme_park: { _score: 2000 },
+    zoo: { _score: 2000 },
+    _score: 1000
+  },
+  museum: { _score: 2000 },
+  museum_type: { _score: 2000 },
+  opening_hours: { _score: 1000 },
+  operator: { _score: 1000 },
+  fee: { _score: 2000 },
+
+  // closed and abandoned places
+  // https://wiki.openstreetmap.org/wiki/Key:disused:
+  disused: { _score: -100000 },
+  'disused:shop': { _score: -100000 },
+  'disused:amenity': { _score: -100000 },
+  'disused:railway': { _score: -100000 },
+  'disused:leisure': { _score: -100000 },
+  // https://wiki.openstreetmap.org/wiki/Key:abandoned:
+  abandoned: { _score: -100000 },
+  'abandoned:shop': { _score: -100000 },
+  'abandoned:amenity': { _score: -100000 },
+  'abandoned:railway': { _score: -100000 },
+  'abandoned:leisure': { _score: -100000 },
+
+  // types of public amenities
+  amenity: {
+    disused: { _score: -100000 },
+    college: { _score: 2000 },
+    library: { _score: 2000 },
+    school: { _score: 1000 },
+    university: { _score: 2000 },
+    bank: { _score: 1000 },
+    clinic: { _score: 1000 },
+    dentist: { _score: 1000 },
+    doctors: { _score: 1000 },
+    hospital: { _score: 2000 },
+    nursing_home: { _score: 1000 },
+    pharmacy: { _score: 1000 },
+    social_facility: { _score: 1000 },
+    veterinary: { _score: 1000 },
+    community_centre: { _score: 1000 },
+    music_venue: { _score: 1000 },
+    nightclub: { _score: 1000 },
+    planetarium: { _score: 1000 },
+    social_centre: { _score: 1000 },
+    theatre: { _score: 1000 },
+    courthouse: { _score: 1000 },
+    coworking_space: { _score: 1000 },
+    dojo: { _score: 1000 },
+    embassy: { _score: 1000 },
+    fire_station: { _score: 1000 },
+    marketplace: { _score: 1000 },
+    place_of_worship: { _score: 1000 },
+    police: { _score: 1000 },
+    post_office: { _score: 1000 },
+    prison: { _score: 1000 },
+    public_bath: { _score: 1000 },
+    townhall: { _score: 1000 }
+  },
+
+  // contact information
+  website: { _score: 2000 },
+  phone: { _score: 1000 },
+  'contact:website': { _score: 2000 },
+  'contact:email': { _score: 1000 },
+  'contact:phone': { _score: 1000 },
+  'contact:fax': { _score: 1000 },
+
+  // social media
+  'contact:foursquare': { _score: 2000 },
+  'contact:facebook': { _score: 2000 },
+  'contact:linkedin': { _score: 2000 },
+  'contact:instagram': { _score: 2000 },
+  'contact:skype': { _score: 2000 },
+  'contact:flickr': { _score: 2000 },
+  'contact:youtube': { _score: 2000 },
+};
+
+module.exports = function(){
+
+  return through.obj(( doc, enc, next ) => {
+
+    try {
+
+      // only map venues
+      if( doc.getLayer() !== 'venue' ){
+        return next(null, doc);
+      }
+
+      // skip records with no tags
+      let tags = doc.getMeta('tags');
+      if( !tags ){
+        return next( null, doc );
+      }
+
+      // default popularity
+      let popularity = doc.getPopularity() || 0;
+
+      // apply scores from config
+      for( let tag in config ){
+        if( tags.hasOwnProperty( tag ) ){
+          // global score for the tag
+          if( config[tag]._score ){
+            popularity += config[tag]._score;
+          }
+          // individual scores for specific values
+          for( let value in config[tag] ){
+            if( value === '_score' ){ continue; }
+            if( !config[tag][value]._score ){ continue; }
+            if( tags[tag] === value.trim() ){
+              popularity += config[tag][value]._score;
+            }
+          }
+        }
+      }
+
+      // set document popularity when it is a non-zero value
+      if( !!popularity ){
+        doc.setPopularity( popularity );
+      }
+    }
+
+    catch( e ){
+      peliasLogger.error( 'popularity_mapper error' );
+      peliasLogger.error( e.stack );
+      peliasLogger.error( JSON.stringify( doc, null, 2 ) );
+    }
+
+    // discard disused and abandoned places with a negative popularity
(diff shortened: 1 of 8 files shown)


bboure:
Thank you @missinglink I am willing to test this. 
I am still quite new to Pelias. If I understand correctly, I would need to re-import OSM data using your branch?

How did you end up with these scores? are they rather arbitrary?
I understand this is very WIP yet, but I suppose these could be configurable?

missinglink:
Yep, it requires reindexing `openstreetmap`, if you are using [pelias/docker](https://github.com/pelias/docker) for your builds then you can test this feature by changing the `api.image` from `pelias/openstreetmap:master` to `pelias/openstreetmap:venue_popularity`.

Once you've done that you can pull the new image [from dockerhub](https://hub.docker.com/r/pelias/openstreetmap/tags) by executing `pelias compose pull`.

Finally you can reimport the openstreetmap data with `pelias import osm` and you should see the results immediately.

The query logic should already be in place to take advantage of updated `popularity` data but it the effect might either be too stong or too weak.
So we may need to either adjust the figures in this config file or adjust the weighting at query-time in order to get the balance right.

The scores are totally arbitrary, I just made them up :)
It will take a bit of trial-and-error during testing to figure out if they are decent or not, once we get it nicely balanced we can commit these values to `master`.

It will also be possible for users to modify the config on their own Pelias setup by editing the file directly, or, using docker, by bind-mounting their own config over the one in the image.

Let me know how you get on

missinglink:
@bboure you may find this useful https://github.com/pelias/docker/pull/103

bboure:
Thank you @missinglink 
I was able to re-import OSM data and made a few tests (I was using Docker)
The effect to be the desired one. Good job! 🎉 
I will keep playing with it and give some feedback.

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+  wikidata: { _score: 3000 },
+
+  // properties found on major landmarks &
+  // public buildings.
+  architect: { _score: 5000 },
+  heritage: { _score: 5000 },
+  'heritage:operator': { _score: 2000 },
+  historic: {
+    building: { _score: 5000 },
+    church: { _score: 5000 },
```
Joxit: historic=monument with a higher score that default?

orangejulius:
Woah, this PR got real big real fast. I really liked the simple initial version, although I still think this is an important thing to add, and really can't believe we haven't done it till now :)

I'm a little hesitant with the many many additions that came in later commits, although I also see how they can be useful. I've been trying to think through ways this PR could cause problems or regressions. I can't come up with much, but I bet it's possibly for a generically named venue (something like "Market") to drown out other more specific results.

It might be worth testing this PR on either a continent or planet scale before merging, to see if we can identify any cases like that, or if there's any unexpected behavior.

missinglink:
One thing that came later was only setting `popularity` for the `venue` layer, if we allowed it on the `address` layer then `osm` would always have a points advantage over `oa`!

I think it's a great thing to have, a couple of hesitations on my part:
- We could possibly remove the 'contact information' scoring completely, I reduced the scoring already but I think that the statement "venues tagged with a phone number are more popular than those without" isn't necessarily true.
- This will need to have some end-to-end testing done before release, the `popularity` scoring subqueries have been in Pelias for a long time but haven't been used since the Quattroshapes days (like 2 years ago). So it's very likely that the balance is off.

Overall I think it's a really nice feature to have and will make the product more professional feeling, we just need to be careful not to upset the balance as a result.

stream/popularity_mapper.js
missinglink: I saw a few cases where some old stone markers at the top of hills were marked `historic:monument` so I'm not so sure about it?

missinglink:
> If it's possible to get the area of the object here, it would be nice to increase the score.
> For exemple, an hospital with a important surface should have an higher score than a point or smaller hospital.
> We can also do this with the type of the object eg `way` versus `node`?

We have bounding boxes for `way` and `relation` records via `pbf2json` and also the [option of](https://github.com/pelias/pbf2json/pull/81) using the `version` metadata to see how often it has been edited in osm.

In the case of a school I think you're right, a big school is probably more important than a small school, however the larger school could be mapped as a `node` and so would appear to be smaller.

The same is also true of monuments, a tower is physically small but could be a more popular tourist attraction than an old football stadium which is much larger.

Total edits could be interesting, it shows at least that the place is popular with mappers and so it's probably important enough to get correct, although I don't know what sort of score we would assign based on how many edits were made.

Thoughts?

bboure:
> Total edits could be interesting, it shows at least that the place is popular with mappers and so it's probably important enough to get correct

I agree on that. A popular place will have many edits in OSM.

I am not sure about the size/are. For example, [Manneken Pis](https://en.wikipedia.org/wiki/Manneken_Pis) is very small, but also very famous 😄 

The number of translations (name.*) might also be a sign that the place is famous internationally? (similar/synonym to `importance=international`)

I have been playing a bit with the current implementation and it gives me pretty accurate results on a City scale map (Barcelona). It probably can improve though. 
I will give more feedback later.

NickStallman:
It might also be worthwhile to have a mechanism to allow for a proportional score.
E.g. start_date is more important the older it is, a date of yesterday and a date 200 years ago aren't equal. Some kind of exponential score here with a upper bound would be good.
And height would be another good one, taller things would often be more important than shorter things.

nvkelso:
🎉 Super exciting progress!

For reference, we added [collision_rank](https://github.com/tilezen/vector-datasource/blob/master/spreadsheets/collision_rank.yaml) in [v1.7.0](https://github.com/tilezen/vector-datasource/releases/tag/v1.7.0) of Tilezen to solve similar "same but different" sort problem.

missinglink:
Heya, nice to see a bunch of interest in this PR!

I've been thinking about this some more over the weekend and I think we all agree that the concept of `popularity` is still pretty vague and not anchored to anything in any logical way that would allow us to estimate the resulting behaviour when we assign popularity values to documents.

In order to deal with that subjectivity, I offer this explanation of `popularity` which we can adopt going forward.

1. Popularity for 'admin' areas is based on their `population`
2. In some admin areas, the popularity will be more or less than the population (such as increasing it in places such as `Venice` or `San Francisco` & decreasing it for large regions and states)
3. All other layers are assigned popularity based on the statement `Assuming there is a city with the same name, assigning higher popularity will rank this place higher than the city, and lower popularity will rank it lower`. This will allow us to, for instance, score `Newark International Airport` higher than `Newark, NJ` in the results.
4. Addresses have zero popularity. We may make some exceptions for things like the `White House`, but this is generally true.
5. Streets have popularity range (still to be decided) which is higher than zero but less than a small city (I'm thinking in the thousands).
6. Venues have the most variable popularity, ranging from `airports` (which may be more popular than their locality) to a low minimum value, such as `100`. As per previous discussions, tourist attractions would be scored in a way that replicas of 'real' places would have lower popularity.

Some advantages of adopting a consistent popularity score across all layers:
- We can score 'real' monuments higher than replicas
- We can increase performance for short autocomplete inputs by filtering on popularity. see https://github.com/pelias/api/pull/1215
- We can better mix layers in results, queries using a `focus.point` will show a better mix of layers based on their relative popularity
- It will be much easier (and bug-free) to exclude addresses from certain queries where they are not relevant.

Thoughts/feels?

missinglink:
Couple notes on elasticsearch scoring based on global popularity values:
- We would no longer need a subquery to score on `population` as `popularity` will serve the same purpose albeit in a more flexible way.
- We will need to 'balance' the effects of the `popularity` bias relative to the textual matching scores. For example 'Newark, NJ' should show the locality higher despite the airport having the largest popularity value.

I think the latter will be easier and more consistent when thee values are more consistent.

missinglink:
I saw Sarah Hoffmann from the [Nominatim Project](https://github.com/openstreetmap/Nominatim) last week and she said their popularity scoring is solely based on Wikipedia.

They compute the 'internal inbound link count' in Wikipedia for each OSM place with a concordance and use that value (ie. 'wiki page rank').

She said they were pretty happy with the results, the dump is [available for download](http://nominatim.org), it's about 6 years old but still pretty relevant. They plan to update the file this year.

bboure:
After this change, would it make sense to add a way in the api to fetch the top `n` venues in a given area?
i.e: I am interested in showing popular places on a map (in a given bbox)

stream/popularity_mapper.js
Joxit: Yes you are right, this is not as relevant as I thought. (I saw a lot of `historic=monument` + `name=*` in Paris and its region.)

bboure:
Hey,
What is the status on this?

> I'd love to see some suggestions from the community on what they would like to see regarding 'importance' scoring in OSM.
> Please suggest additional tags and scoring methodologies!

How about having customizable scoring system?
i.e.: Being able to configure what scope each specific tag should have (or not).

For example, if you are building a transportation system, you might want to boost train/bus stations higher, while still showing other results.
Example with https://www.openstreetmap.org/way/5013364 and https://www.openstreetmap.org/node/5682929172
You would rather show the bus stop first, and then the actual tower. 

Which leads me to thinking that it would also be nice to have a query-time booster by layer or category on the API as well.

`/search?text=Tour+Eiffel&prioritize.categories=transport`

This would give a higher score to documents with the given categories, but would still show other results.

missinglink:
We ([Geocode Earth](http://geocode.earth/)] are currently looking at this issue again and hope to merge some code which allows for improved venue scoring soon.

related: https://github.com/pelias/openstreetmap/issues/385

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+    prison: { _score: 1000 },
+    public_bath: { _score: 1000 },
+    townhall: { _score: 1000 }
+  },
+
+  // transportation
+  aerodrome: {
+    international: { _score: 10000 },
+    regional: { _score: 5000 }
+  },
```
blackmad, line 127: fwiw I found using the small/medium/large categories from https://ourairports.com/data/airports.csv to be more useful than "international" - there are some small and medium international airports that don't deserve such a boost.

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+    post_office: { _score: 1000 },
+    prison: { _score: 1000 },
+    public_bath: { _score: 1000 },
+    townhall: { _score: 1000 }
+  },
+
+  // transportation
+  aerodrome: {
+    international: { _score: 10000 },
+    regional: { _score: 5000 }
```
blackmad, line 126: also maybe want a downweight on aerodrome:type=military https://www.openstreetmap.org/node/369160593 ?

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+  },
+  iata: {
+    _score: 5000,
```
blackmad, line 129: how much will this hurt a query for "CVS" that doesn't want the airport?

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+    church: { _score: 5000 },
+    _score: 1000,
+  },
+  building: {
+    colour: { _score: 2000 },
+    material: { _score: 2000 },
+    supermarket: { _score: 2000 },
+    civic: { _score: 2000 },
+    government: { _score: 2000 },
+    hospital: { _score: 2000 },
```
orangejulius, line 45: add historic here for Wrigley field? https://www.openstreetmap.org/relation/1407988

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+    place_of_worship: { _score: 1000 },
+    police: { _score: 1000 },
+    post_office: { _score: 1000 },
+    prison: { _score: 1000 },
+    public_bath: { _score: 1000 },
+    townhall: { _score: 1000 }
+  },
+
+  // transportation
+  aerodrome: {
```
orangejulius, line 124: Another tag that might make sense is [`aeroway:aerodrome`](https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Daerodrome). It looks like its on most of the major international airports like [Phoenix](https://www.openstreetmap.org/relation/8367318) and [Miami](https://www.openstreetmap.org/way/113657169).

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+    international: { _score: 10000 },
+    regional: { _score: 5000 }
+  },
+  iata: {
+    _score: 5000,
+    none: { _score: -5000 }
+  },
+  railway: {
```
orangejulius, line 132: Looks like another related tag is [`public_transport=station`](https://wiki.openstreetmap.org/wiki/Tag:public%20transport=station?uselang=en-US):

>While the more common tag railway=station is used for all railway stations (i.e. including cargo), the public_transport=station tag is used only on the stations interesting for passenger transport.

This would hopefully help boost some more public transit stops like [Park & Market](https://www.openstreetmap.org/node/6184516702) which currently does not do very well in [our tests](https://github.com/pelias/acceptance-tests/blob/master/test_cases/search_poi.json#L94-L117)

stream/popularity_mapper.js
```diff
@@ -0,0 +1,209 @@
...
+  // concordances
+  wikipedia: { _score: 3000 },
+  wikidata: { _score: 3000 },
+
+  // properties found on major landmarks &
+  // public buildings.
+  architect: { _score: 5000 },
+  heritage: { _score: 5000 },
+  'heritage:operator': { _score: 2000 },
+  historic: {
```
orangejulius, line 34: How about adding [`historic=heritage`](https://wiki.openstreetmap.org/wiki/Tag:historic=heritage?uselang=en-US) for [La Sagrada Familia](https://www.openstreetmap.org/relation/9194723)?

PR #495, opened 2019-06-28
tests(address-extractor) replace event-stream with stream-mock

benetis:
Replacing `event-stream` with `stream-mock` as discussed in https://github.com/pelias/pelias/issues/801


missinglink:
Amazing, thank you!

Issue #497, opened 2019-08-13
Place in OSM not found in Pelias after importing


mihaicosareanu-bolt:
Hi,

I'm trying to import the OSM data for the city of Ganja, Azerbaijan into pelias.

It seems that there are some places that exist in .pbf (I used pbf2json to find them), but they are not present in the elasticsearch index, or I can't seem to find them.

Here is an example of a place: https://www.openstreetmap.org/node/6600732186#map=19/40.68178/46.35879

Even when trying on the following link, I can't seem to find the place: https://geocode.earth/

If you encountered this type of issue before, any advice in how to debug and figure out why it doesn't work would be greatly appreciated :).

missinglink:
We have that restaurant in our production build:
https://pelias.github.io/compare/#/v1/place%3Fids=openstreetmap:address:node/6600732186

Can you please try the path `/v1/place?ids=openstreetmap:address:node/6600732186` against your server to double-check that the document isn't in your index?

missinglink:
It might be that they are present but you aren't able to find them for one reason or another, if that's the case please share some of your queries.

missinglink:
Agh I know what it is actually, the name `"Osmanli Doner"` has been tagged `name:en` when Pelias expects the name to be in the form of another tag, usually `name` (but we also support others).

Could you please check if this is a common theme among your failures?

mihaicosareanu-bolt:
Hmm, I have that entry with the same output. However, the name is missing from the record. Herei s the output of pbf2json: 
```
./build/pbf2json.darwin-x64 -tags='amenity' /tmp/pelias/openstreetmap/424.osm.pbf | grep 6600732186
{"id":6600732186,"type":"node","lat":40.681784,"lon":46.3587859,"tags":{"addr:housenumber":"247","addr:street":"Mustafa Kemal Ataturk","amenity":"restaurant","name:en":"Osmanli Doner","opening_hours":"24/7","phone":"+994502071884","website":"https://www.facebook.com/osmanliganja"}}```

missinglink:
re: https://github.com/pelias/openstreetmap/blob/master/schema/name_osm.js

mihaicosareanu-bolt:
I think I understand. It doesn't seem to have a default name, as others do. Would it be a good solution to make it default to the "name:en" if for some reason "name" doesn't exist?. What do you think?

missinglink:
[...]

PR #498, opened 2019-08-13
use "name:en" as a fallback for when no default name was found

missinglink:
resolves https://github.com/pelias/openstreetmap/issues/497

### DIFF
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -67,6 +67,15 @@ module.exports = function(){
         }
       }
 
+      // Handle the case where no default name was set but there was
+      // an english name which we could use as the default.
+      if( !doc.getName('default') ){
+        var en = doc.getName('en');
+        if( 'string' === typeof en ){
+          doc.setName('default', en);
+        }
+      }
+
       // Import airport codes as aliases
       if( tags.hasOwnProperty('aerodrome') || tags.hasOwnProperty('aeroway') ){
         if( tags.hasOwnProperty('iata') ){
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -67,12 +67,34 @@ module.exports = function(){
         }
       }
 
-      // Handle the case where no default name was set but there was
-      // an english name which we could use as the default.
+      // Handle the case where no default name was set but there were
+      // other names which we could use as the default.
       if( !doc.getName('default') ){
-        var en = doc.getName('en');
-        if( 'string' === typeof en ){
-          doc.setName('default', en);
+
+        var defaultName =
+          doc.getName('official') ||
+          doc.getName('international') ||
+          doc.getName('national') ||
+          doc.getName('regional') ||
+          doc.getName('en');
+
+        // use one of the preferred name tags listed above
+        if ( defaultName ){
+          doc.setName('default', defaultName);
+        }
+
+        // else try to use an available two-letter language name tag
+        else {
+          var keys = Object.keys(doc.name).filter(n => n.length === 2);
+
+          // unambiguous (there is only a single two-letter name tag)
+          if ( keys.length === 1 ){
+            doc.setName('default', doc.getName(keys[0]));
+          }
+
+          // note we do not handle ambiguous cases where the record contains >1
+          // two-letter name tags.
+          // see: https://github.com/pelias/openstreetmap/pull/498
         }
       }
 
--- a/test/stream/tag_mapper.js
+++ b/test/stream/tag_mapper.js
@@ -156,6 +156,25 @@ module.exports.tests.accept_localized_keys = function(test, common) {
   });
 };
 
+// Use 'name:en' in situations where no other default name was available
+// https://github.com/pelias/openstreetmap/issues/497
+module.exports.tests.use_en_as_fallback_default = function (test, common) {
+  var doc = new Document('a', 'b', 1);
+  doc.setMeta('tags', { 'name:ru': 'test1', 'name:pl': 'test2', 'name:en': 'test3' });
+  test('maps - use name:en as fallback default', function (t) {
+    var stream = mapper();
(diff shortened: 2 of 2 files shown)


mihneadb:
@missinglink great stuff for adding this! I was wondering, in the context of https://github.com/pelias/api/pull/1301, wouldn't it make sense to import all available names in the docs?

missinglink:
@mihneadb we're already doing that in https://github.com/pelias/openstreetmap/blob/master/stream/tag_mapper.js#L37-L45, so it's already set up for allowing us to query for various languages at runtime :)

The issue highlighted by https://github.com/pelias/openstreetmap/issues/497 is that the address is available as GID `openstreetmap:address:node/6600732186` but the venue should also be available as  GID`openstreetmap:venue:node/6600732186` too, which it isn't at the moment.

missinglink:
Our code requires a `default` be present or the record is discarded, its set up like that because other code expects a `default` name to always be present.

mihneadb:
Thanks for the insight. That still means that if a place has only a `some-other-language-than-EN` name it will be discarded as a venue, right?

missinglink:
Unfortunately yes, the order of priority for `default` would now be the following OSM tags:

1. name
2. loc_name
3. alt_name
4. short_name
5. name:en

Do you think it would be best to add another level of fallback which accepts any other set language in the case all of these fail?
cc/ @orangejulius thoughts?

missinglink:
We could also check if the `official` or `national` tags are set and use them where available?
ref: https://github.com/pelias/openstreetmap/blob/master/schema/name_osm.js

orangejulius:
This PR looks good as is.

Another thing we might consider is this, for essentially item 6 on the order of priority above:

If there is any single `name:*` tag and no `name` tag, then assign that `name:*` tag to `name.default`

So for example if there is _only_ `name:es`, then that would become the default name.

missinglink:
What about if I change the PR to this:

```
      // Handle the case where no default name was set but there were
      // other names which could use as the default.
      if( !doc.getName('default') ){

        var defaultName =
          doc.getName('official') ||
          doc.getName('international') ||
          doc.getName('national') ||
          doc.getName('regional') ||
          doc.getName('en');

        // use one of the preferred name tags
        if ( defaultName ){
          doc.setName('default', defaultName);
        }
        
        // else just select the first available name tag
        else {
          var tags = Object.keys(doc.name)
          if ( !!tags.length ){
            doc.setName('default', doc.getName(tags[0]));
          }
        }
      }
```

I ran the end-to-end tests and it didn't catch any other cases, but since it's only covering Vancouver (because its small and dual-language), but this issue might be more exaggerated in other parts of the world.

orangejulius:
I don't really like the idea of using the _first_ name tag. That would effectively prioritize languages in alphabetical order, which doesn't really make sense.

It's hard to evaluate the code snippet above, since it's not a diff. How about we merge this PR and then discuss refactoring/other improvements in their own PR?

missinglink:
Yea, picking the first name will either favour tags by lexical sorting or insertion order, which isn't very elegant 😟 

In a rare case where none of the preferred tags are available and `name:en` is also not available and also there are multiple other language names available, in that case, we would pick one at random.

Since its fairly rare (I'm assuming), we might be better of selecting *any* name rather than *no* name, just so the record isn't discarded during import.

Is there possibly another algorithm we could usee to prioritize languages?

mariusa:
> Since its fairly rare (I'm assuming), we might be better off selecting any name rather than no name, just so the record isn't discarded during import.
+1 


> I don't really like the idea of using the first name tag. 
Agreeing that having a name, even first, is better than nothing, let's see if we can improve this.
Note that 'en' is already prioritized. If 'en' isn't present either, what to do?

How about trying the language code which matches the country of the place? This needs a map of official languages for each country code though. It would have to be sorted by language popularity too (eg Switzerland has 4 languages)

missinglink:
okay, here are some stats I generated today from a planet build.

> note: I changed the code slightly to only consider language tags for the 'select first available'
> ie. var keys = Object.keys(doc.name).filter(n => n.length === 2);
> this was to avoid selecting a tag such as 'old_name' which would be undesirable.

Adding the first-level fallback (the one which considers `['official', 'international', 'national', 'regional', 'en']`) would yeild an additional **11587** records globally.

For the second-level fallback (the contentious bit where we select one of the names prefixed with a language code), we could import an additional **2941** records, of which the split is:

```
1x name:** tag  present => 2744 records
2x name:** tags present => 194 records
3x name:** tags present => 3 records
```

some examples:

### 3x name:** fields
- https://www.openstreetmap.org/node/3602069192
- https://www.openstreetmap.org/way/35478443
- https://www.openstreetmap.org/way/625623690

### 2x name:** fields
- https://www.openstreetmap.org/way/593373640
- https://www.openstreetmap.org/way/594675108
- https://www.openstreetmap.org/way/628930715

missinglink:
So I'm going to rebase this PR to solve this issue for the 11587 'simple cases' and also the 2744 'unambiguous' cases.

I'm not going to try and tackle the 197 'ambiguous' cases because I don't have time for it, although I'd be happy if someone else would like to address them.

orangejulius:
Nice analysis. Looks like a really clean and good solution and thankfully the data shows we don't have to make any difficult choices this time :)

stream/tag_mapper.js
```diff
@@ -67,6 +67,37 @@ module.exports = function(){
...
+        if ( defaultName ){
+          doc.setName('default', defaultName);
+        }
+
+        // else try to use an available two-letter language name tag
+        else {
+          var keys = Object.keys(doc.name).filter(n => n.length === 2);
+
+          // unambiguous (there is only a single two-letter name tag)
+          if ( keys.length === 1 ){
```
mihneadb: Thanks for pursuing this! Looks like the best solution.

PR #520, opened 2020-03-19
feat(deps): Remove fs-extra

orangejulius:
We generally only use the fs-extra package for its `mkdir -p` style convenience wrapper, which can now easily be done with the stock fs module.

This removes usage of the fs-extra module and replaces it with native alternatives.

Closes https://github.com/pelias/openstreetmap/pull/519


orangejulius:
Unlike https://github.com/pelias/openaddresses/pull/459, this repo doesn't use anything in `fs-extra` that we can't easily replace! :tada:

PR #528, opened 2020-05-04
Add two new OSM venue tags; add operator/brand to addendum

thismakessand:
Closes #527 

I tested locally using the `portland-metro` project data (504k total records; 36621 venues); after adding brand + operator to the addendum the index size grew by ~0.27% (+~0.3 MB, from 111.95mb -> 112.25mb).

When updating the data used by the end-to-end tests I noticed a couple of new addresses were contained in the output, which I wasn't really expecting...  they appear to be addresses indexed from new venues that were pulled in, eg `relation/6066047`.   I assume this is expected; let me know if not.

Thanks!

PR #529, opened 2020-05-07
additional addendum fields

missinglink:
following on from https://github.com/pelias/openstreetmap/pull/528 this PR uncomments the three fields I wanted to enable in the past but wasn't sure how much additional disk space would be required.

turns out that addendum fields don't add a lot of bytes.

I also took the liberty to modernize the codebase a bit and fix a test case that wasn't being run due to using a duplicate variable name.

### DIFF
--- a/config/features.js
+++ b/config/features.js
@@ -5,12 +5,12 @@
 **/
 
 // default tags imported
-var tags = [
+const tags = [
   'addr:housenumber+addr:street'
 ];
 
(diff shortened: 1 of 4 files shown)


missinglink:
Oh man, I must really want this feature, I did the PR twice!

missinglink:
I cherry-picked the additional tests from the other PR

missinglink:
closes #494

PR #543, opened 2020-10-14
remove language codes we never used

missinglink:
These language codes (`international`, `national`, `old` ... etc) were included in the build years ago and never utilised in query logic.

Let's remove them from the index since they're just taking up space?

note: I had to change the tag mapper slightly to ensure the work from https://github.com/pelias/openstreetmap/pull/498 didn't break.

### DIFF
--- a/schema/name_osm.js
+++ b/schema/name_osm.js
@@ -27,13 +27,15 @@ var OSM_NAMING_SCHEMA = {
   'name':             'default',
   'loc_name':         'default',
   'alt_name':         'default',
-  'int_name':         'international',
-  'nat_name':         'national',
-  'official_name':    'official',
-  'old_name':         'old',
-  'reg_name':         'regional',
   'short_name':       'default',
-  'sorting_name':     'sorting'
+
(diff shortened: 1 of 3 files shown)


orangejulius:
Very nice, this should save us a bit of space in the index.

orangejulius (approved):

Joxit (approved):
`int_name` and `official_name` can be useful in default alias ?

Considering that no one complained, it should be good like this :sweat_smile: .

Edit: oops, too late :sweat_smile:

missinglink:
Oh sorry, I got a bit trigger happy ;)
The main reason was that this shouldn't cause any regressions.

Prior to merging this `int_name` was being stored in `name.international` in the schema, which no-one was using 😆 
So we can have a new PR to add that if it's useful, but this PR shouldn't have removed anything which was being used.

We do actually use `int_name` and `official_name` in the case where `name.default` hasn't been populated from another method, but we don't currently use them as aliases.

Joxit:
Ha ha don't worry :p 

Oh yes you're right they are used as fallback, so it's totally OK :smile:

Issue #507, opened 2019-11-15
duplicate terms in name effecting scoring


missinglink:
It seems as though having duplicate tokens in a name is causing elasticsearch to score the result higher (this is due to how the TF/IDF scoring works)

While it's impossible to have exact duplicate names (this is taken care of by `pelias/model` in a 'post' step), it is possible to have two terms which are very similar such as this:

<img width="289" alt="Screenshot 2019-11-15 at 16 18 39" src="https://user-images.githubusercontent.com/738069/68954248-ee860880-07c3-11ea-88f8-0e02b095d609.png">

https://www.openstreetmap.org/way/432890745

I'm opening this issue so I don't forget, we can either try to solve this during import or during search.

A query such as `/v1/search?text=whole foods market, NY` illustrates the issue (although there may be other things at play here)

<img width="482" alt="Screenshot 2019-11-15 at 16 22 28" src="https://user-images.githubusercontent.com/738069/68954385-32790d80-07c4-11ea-884f-1746c0399672.png">

<img width="477" alt="Screenshot 2019-11-15 at 16 22 16" src="https://user-images.githubusercontent.com/738069/68954349-255c1e80-07c4-11ea-8fbc-5628fa9f0b8a.png">

orangejulius:
Oh yeah, this is a good one. It's kinda inherent to the scoring mechanisms of Elasticsearch.

Maybe we can solve this by modifying the scoring settings of Elasticsearch, which is easier and more flexible in newer versions. We might be able to split altnames into separate fields in a clever way as well. But I think this sort of case will be pretty hard, though not impossible, to solve.

missinglink:
Last night I went into a deep-dive on [tuning the BM25 algorithm](https://www.elastic.co/blog/practical-bm25-part-2-the-bm25-algorithm-and-its-variables), I suspect that we could reduce `k1` to either 0 or near 0 to resolve this issue.

It's a big topic in itself so I'll start a discussion over on `pelias/schema`

orangejulius:
Since we've decided this issue is mostly unsolvable without major changes to how we index data in Elasticsearch, let's talk mitigation:

In this particular case, one name is a prefix of the other. Is it worth it to detect that and skip the shorter version?

In the other main case I've [noticed](https://github.com/pelias/pelias/issues/849) with OSM data, the names only differ by whitespace. That one might be trickier, but we could probably still do it.

PR #548, opened 2021-02-04
Detect altnames that are a substring of name.default

orangejulius:
This change is an attempt to mitigate scoring penalties applied to documents with alternate names (https://github.com/pelias/openstreetmap/issues/507).

It handles the case where an alt name is merely a substring of the main name, for example on the [Union Square](https://www.openstreetmap.org/node/7523716838) subway stop in OSM:

image

Alt names like this don't add much value: they don't allow searching on any new terms, but do throw off the scoring. Even when we fix the scoring issue, duplicate alt names that add no value still take up space, so this change _should_ be useful for a while.

The change comes in 2 parts, each in their own commit:
- what is intended to be a pure [refactoring](https://github.com/pelias/openstreetmap/commit/4678b491c8b59a83b7a343da7155bb3397b3e9e0) of the tag mapping logic. Rather than loop through all tags in whatever order they come, start with the tags that can go into `name.default`, then the other `name.*` fields, and finally the rest. This makes it easier to write logic that looks for duplicates
- the actual change to check for the substring match, which is [quite small](https://github.com/pelias/openstreetmap/commit/e775ef60107ac2411153b320a0f9766f28ebc8ce)

I'd be happy to extend this in the future with other near-identical alt names, such as handling `Mc Donalds` vs `McDonalds`.


test/fixtures/combined_vancouver_queens.json
```diff
@@ -892395,8 +892389,7 @@
     "_type": "_doc",
     "data": {
       "name": {
-        "default": "IPOH Asian House"
-      },
+        "default": "IPOH Asian House" },
```
missinglink, line 892392: non-issue: this is weirdly indented

orangejulius: oh, oops! My fault! I had my hand at editing the fixture manually since there were only a few changes.

Should have left it to the machines.

missinglink: haha, it's really not an issue, I just noticed it a few times in the PR, like the `trim( tags[key]))` catches my eye.

the styling used to be more hit-and-miss, I've been using autoformatting in my editor for a while now which hopefully fixes a bunch of the little ones.

at some point I'd still love to fully adopt [standardJS](https://standardjs.com/).

orangejulius:
I came across this PR today and wanted to see if it still made a difference, so I've rebased it and kicked off a planet build to test things out.

PR #552, opened 2021-03-05
popularity: do not penalize abandoned:amenity=place_of_worship

missinglink:
As per https://github.com/pelias/geonames/pull/394 this PR is to fix the scoring of the query for "Angkor Wat".

Pelias is currently giving the [OSM record](https://www.openstreetmap.org/way/91217761) a negative popularity as it's tagged `abandoned:amenity=place_of_worship`. 

We originally assumed that `abandoned:amenity` meant it was no longer important but that doesn't seem to be the case, either it's mis-tagged or we're interpreting the tag wrong. The [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:abandoned:) seems to confirm our view, but who knows 🤷‍♂️ .

This PR maintains the negative scoring for all `abandoned` and `abandoned:*` tags with a single exception for `abandoned:amenity=place_of_worship`.

Currently that only affects [421 records](https://taginfo.openstreetmap.org/keys/abandoned:amenity#values) so it's unlikely to have any negative side-effects.

### DIFF
--- a/stream/popularity_mapper.js
+++ b/stream/popularity_mapper.js
@@ -82,7 +82,12 @@ const config = {
   // https://wiki.openstreetmap.org/wiki/Key:abandoned:
   abandoned: { _score: -100000 },
   'abandoned:shop': { _score: -100000 },
-  'abandoned:amenity': { _score: -100000 },
+  'abandoned:amenity': {
+    // note: some places such as 'Angkor Wat' are tagged 'abandoned:amenity=place_of_worship'.
+    // we use a positive integer to negate the effects of the base _score below (zeroing it out).
+    place_of_worship: { _score: +100000 },
+    _score: -100000
+  },
   'abandoned:railway': { _score: -100000 },
   'abandoned:leisure': { _score: -100000 },
 
--- a/test/stream/popularity_mapper.js
+++ b/test/stream/popularity_mapper.js
@@ -290,6 +290,20 @@ module.exports.tests.abandoned = function (test, common) {
   });
 };
 
+module.exports.tests.abandoned_amenity_place_of_worship = function (test, common) {
+  var doc = new Document('osm', 'venue', 1);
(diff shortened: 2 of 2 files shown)


missinglink:
Also worth noting that users who enable the config option `imports.openstreetmap.removeDisusedVenues` wpuld have previously had that record omitted from the index, with this change it would be included.

missinglink:
I spent some more time looking at this and TBH it's really not clear how we (and other geocoders) should interpret `abandoned:amenity`.

The [wiki](https://wiki.openstreetmap.org/wiki/Key:abandoned:) says:

> For example, an abandoned shop which still carries a name sign might be tagged as... `abandoned:shop=doityourself`, `name=Holby Tool Supplies` ...which makes it unavailable to programs which locate shops, and unlikely to be displayed as a shop in any default renderings. The name is still relevant, and will still be searchable. If the shop is repaired and reopened, it's a simple matter to undo the prefixing.

Although I disagree with the comment "The name is still relevant, and will still be searchable". I think the author of that comment is specifically referring to it still being returned by nominatim for mappers to find and edit.

It's not clear if Pelias is a "program which locates shops" or a program which makes things "searchable" in that context.

A large number of places tagged `abandoned:amenity` appear to be incorrectly tagged `abandoned` when they should have been tagged `disused`. IMO they shouldn't be considered 'current' or 'active' enough for inclusion in the index, this confusion makes the tag quite unreliable to depend on in any meaningful way.

However, in OSM there's an exception to every rule, there seem to be some places tagged `abandoned:amenity=place_of_worship` which are still culturally relevant, such as [Angkor Wat](https://www.openstreetmap.org/way/91217761) or [Christchurch Cathedral](https://www.openstreetmap.org/way/30366472).

missinglink:
As this change only currently affects 421 records I'm happy to merge it as-is.
I'm open to making changes to how we interpret `abandoned:*` in the future.

Issue #557, opened 2021-08-10
OSM records with 'constructor' tag result in invalid Pelias documents


orangejulius:
Back in https://github.com/pelias/pelias/issues/867 we noticed that our `name.*` dynamic mapping had a strange entry:

In addition to expected fields like `name.default`, `name.en`, etc there was a dynamic name field with the key `function Object() { [native code] }`.

I was looking into language related data today and spent an extra moment to track down what was going on here.

It appears that OSM records with the non-standard [`constructor`](https://taginfo.openstreetmap.org/keys/constructor) tag are the ultimate cause of this odd data.

The good news is there are only 60 records in all of OSM with this tag, and only 32 records ended up with the `function Object() { [native code] }` property in the planet build I was examining.

Still, we probably want to skip the `constructor` tag when importing :)

orangejulius:
Here's all the OSM records that currently cause this issue in our builds:

  'https://openstreetmap.org/way/361486850',
  'https://openstreetmap.org/way/173145217',
  'https://openstreetmap.org/way/43664624',
  'https://openstreetmap.org/way/317549565',
  'https://openstreetmap.org/way/764344502',
  'https://openstreetmap.org/way/770844609',
  'https://openstreetmap.org/way/736843631',
  'https://openstreetmap.org/way/546394548',
  'https://openstreetmap.org/way/666638716',
  'https://openstreetmap.org/way/87685903',
  'https://openstreetmap.org/way/548264277',
  'https://openstreetmap.org/relation/11260075',
  'https://openstreetmap.org/way/546388878',
  'https://openstreetmap.org/way/804562392',
  'https://openstreetmap.org/way/731839615',
  'https://openstreetmap.org/way/360732599',
  'https://openstreetmap.org/way/299651908',
  'https://openstreetmap.org/way/764333216',
  'https://openstreetmap.org/node/281226115',
  'https://openstreetmap.org/way/770844629',
  'https://openstreetmap.org/way/831261068',
  'https://openstreetmap.org/way/44063283',
  'https://openstreetmap.org/way/597564764',
  'https://openstreetmap.org/way/736843642',
  'https://openstreetmap.org/way/548280145',
  'https://openstreetmap.org/way/548262350',
  'https://openstreetmap.org/way/315890958',
  'https://openstreetmap.org/way/547581846',
  'https://openstreetmap.org/way/817810785',
  'https://openstreetmap.org/way/548268061',
  'https://openstreetmap.org/way/547579822',
[...]

PR #558, opened 2021-08-10
tag_mapper: skip inherited enumerable properties

missinglink:
I was interested in https://github.com/pelias/openstreetmap/issues/557, it seems that we are using the [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) / [for ... in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) syntax which iterates over all properties "including inherited enumerable properties".

A quick `git grep` shows other usages of that syntax in this lib which might need attention too.

closes https://github.com/pelias/openstreetmap/issues/557

### DIFF
--- a/config/category_map.js
+++ b/config/category_map.js
@@ -17,7 +17,7 @@
  @see: https://github.com/pelias/pelias/wiki/Taxonomy-v1
 **/
 
-var mapping = {
+const mapping = {
 
   'aerialway': {
     '*':                        ['transport'],
--- a/config/localized_name_keys.js
+++ b/config/localized_name_keys.js
@@ -30,14 +30,15 @@
   @see: http://wiki.openstreetmap.org/wiki/Talk:Names
 **/
 
-var iso6393 = require('iso-639-3');
-var keys = [];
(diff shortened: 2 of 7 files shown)


orangejulius:
Nice, I haven't touched this code in a while so I wasn't quite sure where the fix would end up. Looks like the right place. If you think any of the other `for ... in` code needs fixing, go for it. Otherwise this looks good to merge 🚢

missinglink:
alrighty, not the most enjoyable task of the week 😆 but I've added https://github.com/pelias/openstreetmap/pull/558/commits/36bfd97e75b4db5a44363ab29b8115e4a7923565 which replaces all usages of the [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) / [for ... in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) syntax.

the nice thing about this repo is we have pretty decent unit test coverage and the `end-to-end` tests which helped me resolve any issues.

PR #560, opened 2021-08-20
tag_mapper: contract english diagonals for street addresses

missinglink:
Some addresses are failing to deduplicate because of minor differences in the expansion/contraction of diagonals:

Screenshot 2021-08-20 at 12 45 00

This PR is very basic, it simply replaces the expanded form of these diagonals (in English only) where found in a street field.

We have a more advanced strategy for `openaddresses` in https://github.com/pelias/openaddresses/pull/477 which I'd like to port over to this repo too at some point, one difficulty with that is detecting the language appropriately.

This PR is a very simple fix in the interim.

The `pelias/schema` codebase has [synonym mappings](https://github.com/pelias/schema/blob/7eca249ea3657f08d4e758c5351a18b3ec79df8f/synonyms/directionals/en.txt#L16) which will allow matching to work in both forms.

### DIFF
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -62,7 +62,8 @@ module.exports = function(){
         else if( _.has(ADDRESS_SCHEMA, key) ){
           var val3 = trim( value );
           if( val3 ){
-            doc.setAddress( ADDRESS_SCHEMA[key], val3 );
+            let label = ADDRESS_SCHEMA[key];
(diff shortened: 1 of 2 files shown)


orangejulius (approved):
Oh yeah, good stuff. I'd been thinking about this ever since we improved the abbreviation handling in OpenAddresses. It will be great for them to be consistent.

PR #565, opened 2022-03-09
Added missing street mapping

Bimbol:
Because of missing tag `addr:place` in schema, importer skip a lot of addresses. For example in Mannheim, DE most of the streets named with letter is not imported.
image

So to fix this, I added missing tag.

### DIFF
--- a/stream/addresses_without_street.js
+++ b/stream/addresses_without_street.js
@@ -24,14 +24,17 @@ module.exports = function(){
         return next( null, doc );
       }
 
-      // fill addr:street with addr:place if missing
-      var street = tags['addr:street'] || '';
-      if (
-        street === '' &&  // override also if street tag is empty
-        tags.hasOwnProperty('addr:place') &&
-        tags.hasOwnProperty('addr:housenumber')
-      ) {
-        tags['addr:street'] = tags['addr:place'];
+      // housenumber is required
+      if (!_.has(tags, 'addr:housenumber')){
+        return next( null, doc );
+      }
+
+      const street = _.get(tags, 'addr:street', '').trim();
+      const place = _.get(tags, 'addr:place', '').trim();
+
+      // when street is unset but place is set, use place for the street name
+      if (_.isEmpty(street) && !_.isEmpty(place)) {
+        _.set(tags, 'addr:street', place);
       }
     }
 
--- a/stream/addresses_without_street.js
+++ b/stream/addresses_without_street.js
@@ -25,13 +25,13 @@ module.exports = function(){
       }
 
       // fill addr:street with addr:place if missing
-      var street = tags['addr:street'] || ''
+      var street = tags['addr:street'] || '';
       if (
-        street == '' &&  // override also if street tag is empty
+        street === '' &&  // override also if street tag is empty
         tags.hasOwnProperty('addr:place') &&
         tags.hasOwnProperty('addr:housenumber')
       ) {
-        tags['addr:street'] = tags['addr:place']
+        tags['addr:street'] = tags['addr:place'];
       }
     }
 
--- a/stream/addresses_without_street.js
+++ b/stream/addresses_without_street.js
@@ -1,0 +1,52 @@
+
+/**
+  This stream is responsible for filling 'addr:street' in case if
+  is empty or missing, with 'addr:place' value. Note that both 'addr:place'
+  and 'addr:housenumber' must be present inside tags!
+
+  @ref https://github.com/pelias/openstreetmap/pull/565
+**/
+
+const _ = require('lodash');
+const through = require('through2');
+const peliasLogger = require('pelias-logger').get('openstreetmap');
+
+
+module.exports = function(){
+
+  var stream = through.obj( function( doc, enc, next ) {
+
+    try {
+
+      // skip records with no tags
+      var tags = doc.getMeta('tags');
+      if( !tags ){
+        return next( null, doc );
+      }
+
+      // fill addr:street with addr:place if missing
+      var street = tags['addr:street'] || ''
+      if (
+        street == '' &&  // override also if street tag is empty
+        tags.hasOwnProperty('addr:place') &&
+        tags.hasOwnProperty('addr:housenumber')
+      ) {
+        tags['addr:street'] = tags['addr:place']
+      }
+    }
+
+    catch( e ){
+      peliasLogger.error( 'addresses_without_street error' );
+      peliasLogger.error( e.stack );
+      peliasLogger.error( JSON.stringify( doc, null, 2 ) );
+    }
+
+    return next( null, doc );
+
+  });
+
+  // catch stream errors
+  stream.on( 'error', peliasLogger.error.bind( peliasLogger, __filename ) );
+
+  return stream;
+};
--- a/stream/addresses_without_street.js
+++ b/stream/addresses_without_street.js
@@ -1,8 +1,8 @@
 
 /**
-  This stream is responsible for filling 'addr:street' in case if
-  is empty or missing, with 'addr:place' value. Note that both 'addr:place'
-  and 'addr:housenumber' must be present inside tags!
+  This stream is responsible for filling 'addr:street' if is empty or missing,
+  with 'addr:place' value. Note that both 'addr:place' and 'addr:housenumber'
+  must be present inside tags!
 
   @ref https://github.com/pelias/openstreetmap/pull/565
 **/
--- a/test/stream/addresses_without_street.js
+++ b/test/stream/addresses_without_street.js
@@ -29,7 +29,7 @@ module.exports.tests.passthrough = function(test, common) {
     original.setMeta('tags', { 'addr:place': 'L14' });
     var stream = mapper();
     stream.pipe( through.obj( function( doc, enc, next ){
-      t.deepEqual(doc.getMeta('tags'), original.getMeta('tags'), 'tags modified' );
+      t.deepEqual(doc.getMeta('tags'), original.getMeta('tags'), 'tags not modified' );
       t.end(); // test will fail if not called (or called twice).
       next();
     }));
--- a/test/stream/addresses_without_street.js
+++ b/test/stream/addresses_without_street.js
@@ -1,0 +1,102 @@
+
+const through = require('through2');
+const mapper = require('../../stream/addresses_without_street');
+const Document = require('pelias-model').Document;
+
+module.exports.tests = {};
+
+// test exports
+module.exports.tests.interface = function(test, common) {
+  test('interface: factory', function(t) {
+    t.equal(typeof mapper, 'function', 'stream factory');
+    t.end();
+  });
+  test('interface: stream', function(t) {
+    var stream = mapper();
+    t.equal(typeof stream, 'object', 'valid stream');
+    t.equal(typeof stream._read, 'function', 'valid readable');
+    t.equal(typeof stream._write, 'function', 'valid writeable');
+    t.end();
+  });
+};
+
+// a tags with missing of required extra tag (addr:housenumber)
+// should pass through the stream without being modified.
+// @ref https://github.com/pelias/openstreetmap/pull/565#issuecomment-1062874227
+module.exports.tests.passthrough = function(test, common) {
+  test('passthrough: missing addr:housenumber tag', function(t) {
+    var original = new Document('a','b', 1);
+    original.setMeta('tags', { 'addr:place': 'L14' });
+    var stream = mapper();
+    stream.pipe( through.obj( function( doc, enc, next ){
+      t.deepEqual(doc.getMeta('tags'), original.getMeta('tags'), 'tags modified' );
+      t.end(); // test will fail if not called (or called twice).
+      next();
+    }));
+    stream.write(original);
+  });
+};
+
+// // ======================== addresses =========================
+
+module.exports.tests.karlsruhe_schema = function(test, common) {
+  test('maps - karlsruhe schema with filled street', function(t) {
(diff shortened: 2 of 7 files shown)


missinglink:
Hi @Bimbol, we've had a few reports about these unusual addressing schemas over the years, for instance in Polish towns you'll find the use of a `housenumber` but combined with the locality name instead of the `street`, in this case the grid reference is used in place of the `street` name.

In both cases it seems the preferred way of mapping it in OSM is using the `addr:place` tag, which is nice, but this tag is also commonly used for addresses mapped with the more conventional syntax too.

The proposal in your PR would set an `addr:street` when available, but then replace it with `addr:place` if *also* available, this is not correct as `addr:street` is preferred when available.

Other than that, it should, to some degree 'just work' with the place name being used in the street field, of course the parsers are going to get confused but hopefully it's handled fairly gracefully.

tl;dr, these are 'special cases', so we need to take care to 'special case' them correctly.

missinglink:
Also worth noting that the [Karlsruhe Schema](https://wiki.openstreetmap.org/wiki/Proposed_features/House_numbers/Karlsruhe_Schema) has defined meaning, I don't believe these addressing schemas are compatible, so.. while convenient, it might make more sense to have this in a new file with some comments about what it is and maybe a link or two for future reference.

missinglink:
for reference here is an address from Mannheim with this addressing schema:
- https://www.openstreetmap.org/way/320826090

here's an example from the same area where using the `addr:place` when a `addr:street` is available is incorrect:
- https://www.openstreetmap.org/relation/4134026

discussion in German:
- https://forum.openstreetmap.org/viewtopic.php?id=62139

missinglink (changes requested):
The code must prefer `addr:street` over `addr:place` when available.

Bimbol:
Thank you for explanation. I was wondering why this file has city name in it :)

So changing the order of schemas in merge could resolve issue with priority I guess?

missinglink:
oh also worth mentioning that the config here: https://github.com/pelias/openstreetmap/blob/master/config/features.js#L9 will need to change, these settings control which tags are extracted by [pbf2json](https://github.com/pelias/pbf2json), you'll need to add another line below the one I linked with says `'addr:housenumber+addr:place'`.

orangejulius:
Hey @Bimbol thanks for bringing this up, we should definitely support `addr:place` where possible.

I can confirm that: `addr:street` and `addr:place` are not supposed to both be set, according to the [OSM wiki page for `addr:place](https://wiki.openstreetmap.org/wiki/Key:addr:place). There's also some discussion about how many uses of `addr:place`, especially in the UK, [don't follow that rule](https://github.com/openstreetmap/id-tagging-schema/issues/353#issuecomment-1038010108)

But like @missinglink mentioned in a previous discussion, we may not want to edit this particular file to do it: https://github.com/pelias/pelias/issues/787#issuecomment-477137803. Is that still good advice @missinglink? I don't want to create _too_ much work but I can see the need for an `addr:place` specific schema.

missinglink:
> So changing the order of schemas in merge could resolve issue with priority I guess?

Yeah, there's a bunch of ways you could code it up, but maybe just splitting it out to its own file and then finding a way to get it to only apply when `addr:street` isn't present but `addr:housenumber` & `addr:place` are set would work.

Unit tests please!

Bimbol:
Can I modify tags inside pipeline, or should the original ones be kept? Because one of solution is remove "invalid" tag `addr:place` when `addr:street` is present.

missinglink:
I guess you can edit the tags, although it might make debugging more difficult.

So for instance you *could* create a new stream and add it in *before* `address_extractor.js`, call it `addresses_without_street.js`.

In that steam simply check the tags for the existence of `addr:housenumber AND addr:place AND NOT addr:street`, in this situation just copy the `addr:place` to `addr:street`.

It would be easy to write tests for since it would be totally contained and the rest of the code would 'just work' without modification.

This comment is still required: https://github.com/pelias/openstreetmap/pull/565#issuecomment-1062872146

reference: https://help.openstreetmap.org/questions/44160/addresses-with-no-street-names

Bimbol:
> I guess you can edit the tags, although it might make debugging more difficult.
> 
> So for instance you _could_ create a new stream and add it in _before_ `address_extractor.js`, call it `addresses_without_street.js`.
> 
> In that stream simply check the tags for the existence of `addr:housenumber AND addr:place AND NOT addr:street`, in this situation just copy the `addr:place` to `addr:street`.
> 
> It would be easy to write tests for since it would be totally contained and the rest of the code would 'just work' without modification or risk of introducing failing cases which would make it hard to approve for merging.
> 
> This comment is still required: [#565 (comment)](https://github.com/pelias/openstreetmap/pull/565#issuecomment-1062872146)
> 
> reference: https://help.openstreetmap.org/questions/44160/addresses-with-no-street-names

I applied some changes according to your suggestions. I am waiting for some feedback :)
I was also thinking if this could be configured somehow, however simple configuration may not cover all potencial use cases.

Thank you for your time!

test/stream/addresses_without_street.js
```diff
@@ -0,0 +1,102 @@
...
+// a tags with missing required extra tag (addr:housenumber)
+// should pass through the stream without being modified.
+// @ref https://github.com/pelias/openstreetmap/pull/565#issuecomment-1062874227
+module.exports.tests.passthrough = function(test, common) {
+  test('passthrough: missing addr:housenumber tag', function(t) {
+    var original = new Document('a','b', 1);
+    original.setMeta('tags', { 'addr:place': 'L14' });
+    var stream = mapper();
+    stream.pipe( through.obj( function( doc, enc, next ){
+      t.deepEqual(doc.getMeta('tags'), original.getMeta('tags'), 'tags modified' );
```
missinglink, line 32: nit: 'tags modified' should probably say 'tags not modified'?

stream/addresses_without_street.js
```diff
@@ -0,0 +1,52 @@
...
+
+      // fill addr:street with addr:place if missing
+      var street = tags['addr:street'] || '';
+      if (
+        street === '' &&  // override also if street tag is empty
+        tags.hasOwnProperty('addr:place') &&
+        tags.hasOwnProperty('addr:housenumber')
+      ) {
+        tags['addr:street'] = tags['addr:place'];
+      }
```
missinglink, line 35: This is really great, I just wonder if could just be a bit easier to read 🤔 

we could start with the house number requirement, if this isn't set then it's a no-op anyway:

```
// housenumber is required
if (!_.has(tags, 'addr:housenumber')){ return next( null, doc ); }
```

and then *both* strings should be checked for existence and length:

```
const street = _.get(tags, 'addr:street', '').trim();
const place = _.get(tags, 'addr:place', '').trim();

// when street is unset but place is set, use place for the street name
if( _.isEmpty(street) && !_.isEmpty(place) ) {
 _.set(tags, 'addr:street', place);
}
```

missinglink (approved):
Really great stuff thanks.

I added some minor feedback as a code review.

Did you get a chance to test this? maybe we add a [pelias/docker](https://github.com/pelias/docker) project for Mannheim?

missinglink:
Feedback resolved via new commit, the code in this PR looks good-to-merge.

I'd like to just confirm that it solves the reported issue, from what I can see it's unlikely to introduce error to existing results, which is great, but I wanted to make sure it *actually* fixes the issue in Mannheim (or at least gets us closer to a fix).

Bimbol:
Sorry, that I didn't reply after committing changes. I was still running some tests. Everything seems to work fine. I did checked quite a few streets, that were missing :)
Zrzut ekranu z 2022-03-14 10-14-00

Bimbol:
> Feedback resolved via new commit, the code in this PR looks good-to-merge.
> 
> I'd like to just confirm that it solves the reported issue, from what I can see it's unlikely to introduce error to existing results, which is great, but I wanted to make sure it _actually_ fixes the issue in Mannheim (or at least gets us closer to a fix).

Yes, it resolves the issue with missing addresses in Mannheim.

Bimbol:
@missinglink Is still any problem with this PR? Because is open for a while now.

missinglink:
There are a lot of organisations running Pelias in production so we always test global changes such as this on a wider scale before merging.

This is mostly to ensure no regressions are introduced, but also to ensure that the proposed solution adequately resolves the reported problem, there are often unforeseen side effects exposed which aren't always apparent during dev work, such is the nature of a search engine, you can't rely only on unit tests.

I've made a copy of your `master` branch in this org so you're free to use it for something else, if you'd like to potentially add to this feature then I'd recommend you cut a dev branch on your fork and open the PR from there.

Since it's now in our org you'll find Docker images are automatically generated, one for the branch and one for the commit:
- [pelias/openstreetmap:addresses_without_street](https://hub.docker.com/layers/pelias/openstreetmap/addresses_without_street/images/sha256-550bda355cb78c52d2aa7d75ef586953254e4f67ce52202e77c4b08b85887d9e?context=explore)
- [addresses_without_street-2022-03-16-379984c2543ead806f9d930784a81047e0f2c933](https://hub.docker.com/layers/pelias/openstreetmap/addresses_without_street-2022-03-16-379984c2543ead806f9d930784a81047e0f2c933/images/sha256-550bda355cb78c52d2aa7d75ef586953254e4f67ce52202e77c4b08b85887d9e?context=explore)

I'll kick off a global build using the commit above and run the [acceptance-tests](https://github.com/pelias/acceptance-tests/tree/master/test_cases) to test for any unforseen issues.

Bimbol:
> There are a lot of organisations running Pelias in production so we always test global changes such as this on a wider scale before merging.
> 
> This is mostly to ensure no regressions are introduced, but also to ensure that the proposed solution adequately resolves the reported problem, there are often unforeseen side effects exposed which aren't always apparent during dev work, such is the nature of a search engine, you can't rely only on unit tests.
> 
> I've made a copy of your `master` branch in this org so you're free to use it for something else, if you'd like to potentially add to this feature then I'd recommend you cut a dev branch on your fork and open the PR from there.
> 
> Since it's now in our org you'll find Docker images are automatically generated, one for the branch and one for the commit:
> 
> * [pelias/openstreetmap:addresses_without_street](https://hub.docker.com/layers/pelias/openstreetmap/addresses_without_street/images/sha256-550bda355cb78c52d2aa7d75ef586953254e4f67ce52202e77c4b08b85887d9e?context=explore)
> * [addresses_without_street-2022-03-16-379984c2543ead806f9d930784a81047e0f2c933](https://hub.docker.com/layers/pelias/openstreetmap/addresses_without_street-2022-03-16-379984c2543ead806f9d930784a81047e0f2c933/images/sha256-550bda355cb78c52d2aa7d75ef586953254e4f67ce52202e77c4b08b85887d9e?context=explore)
> 
> I'll kick off a global build using the commit above and run the [acceptance-tests](https://github.com/pelias/acceptance-tests/tree/master/test_cases) to test for any unforseen issues.

Those images are exactly what I need right now. So thank you for this great project :)

missinglink:
This feature is up on the Geocode Earth dev server right now (although be aware that server changes branches we're testing often).

##### Seems to work great for addresses in Mannheim:

https://pelias.github.io/compare/#/v1/autocomplete?text=8+Q1+mannheim

Screenshot 2022-03-17 at 13 28 51

##### Doesn't work for streets in Mannheim:

https://pelias.github.io/compare/#/v1/autocomplete?text=C4%2C+mannheim

This would require a corresponding patch to the commands we use to generate [polylines](https://github.com/pelias/polylines) extracts, so *both* my `missinglink/pbf street` command and the valhalla routing engine we advocate using for production builds.

##### Unfortunately doesn't solve https://github.com/pelias/pelias/issues/642

https://pelias.github.io/compare/#/v1/autocomplete?text=Doln%C3%AD+%C4%8Cermn%C3%A1+314%2C+CZ

This is due to the parser identifying that the locality name is in fact a locality and so not querying the street field, we **might** be able to do something about that but it would need some consideration.

Bimbol:
> ##### Doesn't work for streets in Mannheim:
> https://pelias.github.io/compare/#/v1/autocomplete?text=C4%2C+mannheim
> 
> This would require a corresponding patch to the commands we use to generate [polylines](https://github.com/pelias/polylines) extracts, so _both_ my `missinglink/pbf streets` command and the valhalla routing engine we advocate using for production builds.

I did imported whole Germany today and in my case _search_ works great, but as in example _autocomplete_ not.
Here are some sample data from my database, for input `C4, Mannheim`. I don't know why but with the same input for your dev server I get only cities in result.
```json
"places": [
    {
      "gid": "pelias:address:way/197559108",
      "label": "C4 8, Mannheim, BW, Niemcy",
      ...
    },
    {
      "gid": "pelias:address:way/130369581",
      "label": "C4 9b, Mannheim, BW, Niemcy",
      ...
    },
    {
      "gid": "pelias:address:way/181346505",
      "label": "C4 1, Mannheim, BW, Niemcy",
      ...
    },
    ...
]
```

About the issue you you mentioned, I'm not quite sure if I understand whats the problem here. Could you please give me some more details?

missinglink:
Okay this all looks good to me. Globally, it pulls in an additional ~9MM addresses, which is more than I expected.

There was an issue with the deployment yesterday where the dev machine ran out of disk, so it wasn't quite fully loaded, we resolved that issue and it seems the behaviour of `/v1/search` now matches what you were seeing on your local installation:

https://pelias.github.io/compare/#/v1/search?sources=osm&layers=address&text=C4%2C+mannheim

There is still additional work required to adapt the queries for this new format, but for now this is GTG 🎉

missinglink:
thanks @Bimbol the robots will cut new docker images and tag them `latest` and `master` now, they'll be available in a couple minutes.

PR #567, opened 2022-04-01
tags: map alternate names to individual fields

missinglink:
partner PR for https://github.com/pelias/api/pull/1620


missinglink:
I've kicked off a full planet build @additional-name-fields-2022-04-01-55f13ca329927bcfa31a576504b9920801ee9046

missinglink:
this looks promising, grepping the fixture file shows that *all* name fields now only contain a single entry

orangejulius:
Yeah, the POI test cases for major landmarks look very good with this change.

Just for some context, here's some relevant tests "before":
image

And "after" the changes as of around April 5th:
image

Note that the percentage pass rate shows an improvement only from 84 to 86%, but the actual increase is a lot higher, since most of the tests are expected to fail. In actuality, there's a net increase of 25 new requests passing, or a 6% improvement. Considering there are a few regressions that might either be a testing artifact or something we can fix, and the nature of many of these autocomplete queries meaning many of the shorter test cases won't ever pass, that's pretty good.

PR #578, opened 2024-03-13
support easticsearch v8

missinglink:
this PR modifies the importer such that it works for both elasticsearch v7 & v8
in order to support v8 this PR drops support for v6.


missinglink:
This is the final image compatible with elasticsearch v6:
https://hub.docker.com/layers/pelias/openstreetmap/v5.20.0/images/sha256-745d0d62635d0e2c72499b99c7b8fc342ed17bac996a068e5057d90d6ade3b64?context=explore

Issue #51, opened 2015-04-09
Multiple values for same OSM tag key


missinglink:
Ensure we are handing 'semi-colon value separators' correctly.

http://wiki.openstreetmap.org/wiki/Talk:Semi-colon_value_separator

missinglink:
I spoke with Jochen Topf about this, so I have a better idea of how to dev this now.

missinglink:
related: https://github.com/pelias/openstreetmap/issues/21

missinglink:
This issue has been resolved for `addr:housenumber` in https://github.com/pelias/openstreetmap/issues/21

Implementing this for all properties is not realistic at this point, so I'm marking this ticket as `wontfix`, feel free to comment on it in the future if a specific data error causes frustration.

PR #581, opened 2024-11-29
Properly process alt_name tag when there are multiple entries separated by semicolon

SiarheiFedartsou:
:wave: I did some awesome work for the Pelias project and would love for everyone to have a look at it and provide feedback.

---
#### Here's the reason for this change :rocket:

OSM's alt_name tag may contain multiple entries which are separated by semicolon (`;`) that's why I propose to split by it and use each entry as separate alias name. 
See: https://wiki.openstreetmap.org/wiki/Key:alt_name#Examples
Example: https://www.openstreetmap.org/way/31151623

Screenshot 2024-11-29 at 14 49 26

---
#### Here's what actually got changed :clap:
Just additional `split` call for `alt_name` tags :) 

---
#### Here's how others can test the changes :eyes:
I added new test (which is almost copy-paste of one of existing ones though, but couldn't come up with better solution)

### DIFF
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -50,7 +50,16 @@ module.exports = function(){
             if( key === NAME_SCHEMA._primary ){
               doc.setName( NAME_SCHEMA[key], val2 );
             } else if ( 'default' === NAME_SCHEMA[key] ) {
-              doc.setNameAlias( NAME_SCHEMA[key], val2 );
+              // `alt_name` may contain multiple names separated by semicolons 
+              // (see https://wiki.openstreetmap.org/wiki/Key:alt_name#Examples)
+              if (key === 'alt_name') {
+                const altNames = val2.split(';');
+                altNames.forEach((altName) => {
+                  doc.setNameAlias( NAME_SCHEMA[key], altName );
+                });
+              } else {
+                doc.setNameAlias( NAME_SCHEMA[key], val2 );
+              }
             } else {
               doc.setName( NAME_SCHEMA[key], val2 );
             }
--- a/test/stream/tag_mapper.js
+++ b/test/stream/tag_mapper.js
@@ -108,6 +108,33 @@ module.exports.tests.osm_names = function(test, common) {
       t.end(); // test will fail if not called (or called twice).
       next();
     }));
+    
+    stream.write(doc);
+  });
+
+  test('maps - name aliases - multiple alt_names', function(t) {
+    var doc = new Document('a','b',1);
+    doc.setMeta('tags', {
+      loc_name: 'loc_name',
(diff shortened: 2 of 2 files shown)


stream/tag_mapper.js
```diff
@@ -50,7 +50,16 @@ module.exports = function(){
             if( key === NAME_SCHEMA._primary ){
               doc.setName( NAME_SCHEMA[key], val2 );
             } else if ( 'default' === NAME_SCHEMA[key] ) {
-              doc.setNameAlias( NAME_SCHEMA[key], val2 );
+              // `alt_name` may contain multiple names separated by semicolons 
+              // (see https://wiki.openstreetmap.org/wiki/Key:alt_name#Examples)
+              if (key === 'alt_name') {
```
SiarheiFedartsou, line 55: Tbh at glance it seems to be better to "bake" the fact that we need to split by something to `NAME_SCHEMA`, but it seems to be a bit of over-engineering for now (but let me know :) )

missinglink:
We're doing something similar already for addresses FYI:

https://github.com/pelias/openstreetmap/blob/75309fa6b502d90cdbe4af33e6dc008a6bc224f9/stream/address_extractor.js#L48-L50

missinglink:
Seems this day has come 😆 
https://github.com/pelias/openstreetmap/issues/51#issuecomment-101684508

SiarheiFedartsou:
> We're doing something similar already for addresses FYI:

Do you think I should change this PR somehow to be aligned with this implementation? :)

SiarheiFedartsou:
Btw it seems `alt_name:XXX` (e.g. `alt_name:en` or `alt_name:th`) are not supported, or I am missing something?

missinglink:
> Btw it seems `alt_name:XXX` (e.g. `alt_name:en` or `alt_name:th`) are not supported, or I am missing something?

For alt specifically? If so let's please open a new issue so we can discuss separately

SiarheiFedartsou:
> > Btw it seems `alt_name:XXX` (e.g. `alt_name:en` or `alt_name:th`) are not supported, or I am missing something?
> 
> For alt specifically? If so let's please open a new issue so we can discuss separately

Yep, please see https://github.com/pelias/openstreetmap/issues/583

SiarheiFedartsou:
Hey @missinglink 
What do you think can be changed here to eventually merge it to `master`? :)

orangejulius:
This seems like a good change to me, agree we can merge it. Thanks 🙏

orangejulius (approved):

missinglink:
Sorry I had this on my to-do list for today but didn't get to it, I'll try to merge it Monday, I'd like to quickly look over it again at my computer.

PR #589, opened 2025-02-24
support semi-colon delimited names

missinglink:
replaces https://github.com/pelias/openstreetmap/pull/581

> note: commits need to be squashed and correctly prefixed (with `feat:`) before merging.

This PR adds support for semi-colon delimited names in OSM.
see: https://wiki.openstreetmap.org/wiki/Talk:Semi-colon_value_separator

Compared to #581 I added:
- Cleaning parsed values (trimming whitespace and filtering empty strings, ie. `''` or `foo;;bar`)
- Support for semi-colon delimited names in lang fields (ie. `name:en`) as well as those in `NAME_SCHEMA`.
- Moved `parseSemicolonDelimitedValues()` function to it's own module as it's used in the `address_extractor` too.
- Removed this weird `_primary` key from `NAME_SCHEMA` which probably seemed like a good idea at the time but is unnecessarily confusing.
- Updated the code which hunts for a default name when one wasn't available, also made this semi-colon aware.
- Code modernisation & cleanup - variables were confusingly named, code was dated (using `var` instead of `const` etc.)

closes https://github.com/pelias/openstreetmap/pull/581

### DIFF
--- a/schema/name_osm.js
+++ b/schema/name_osm.js
@@ -14,16 +14,13 @@
   When multiple keys have the value 'default' then they are considered
   as aliases of the default field.
 
-  The '_primary' property defined below defines which of those aliases
-  is considered the 'primary default name' for label generation.
-
   No values other than 'default' should be specified more than once.
 
   @ref: http://wiki.openstreetmap.org/wiki/Key:name
   @ref: http://wiki.openstreetmap.org/wiki/Names
 **/
 
-var OSM_NAMING_SCHEMA = {
+const OSM_NAMING_SCHEMA = {
   'name':             'default',
   'loc_name':         'default',
   'alt_name':         'default',
@@ -37,15 +34,5 @@ var OSM_NAMING_SCHEMA = {
   // 'reg_name':         'regional',
   // 'sorting_name':     'sorting'
 };
-
-// this property is considered the 'primary name'
-// for label generation, the others are considered
-// secondary or 'aliases'.
-Object.defineProperty(OSM_NAMING_SCHEMA, '_primary', {
-  value: 'name',
-  enumerable: false,
-  configurable: false,
-  writable: false
-});
 
 module.exports = OSM_NAMING_SCHEMA;
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -47,7 +47,7 @@ module.exports = function(){
         else if( _.has(NAME_SCHEMA, key) ){
           const nameValue = trim( value );
           if( nameValue ){
-            if( key === NAME_SCHEMA._primary ){
+            if( 'name' === key ){
               doc.setName( NAME_SCHEMA[key], nameValue );
             } else if ( 'default' === NAME_SCHEMA[key] ) {
               doc.setNameAlias( NAME_SCHEMA[key], nameValue );
--- a/stream/tag_mapper.js
+++ b/stream/tag_mapper.js
@@ -7,6 +7,7 @@
 const _ = require('lodash');
 const through = require('through2');
 const peliasLogger = require('pelias-logger').get('openstreetmap');
+const parseSemicolonDelimitedValues = require('../util/parseSemicolonDelimitedValues');
 
 const LOCALIZED_NAME_KEYS = require('../config/localized_name_keys');
 const NAME_SCHEMA = require('../schema/name_osm');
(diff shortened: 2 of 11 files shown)


missinglink:
There are some confusing aspects of this code, I probably wrote a bunch of this, but revisiting it after so long I felt it wasn't super intuitive, I made some attempts to improve it.

Some things which are still a bit clunky are:
- The way `doc.setName()` and `doc.setNameAlias()` work is kinda clunky, it requires the caller to know if that specific name property was previously set before selecting a method to use. It would be much easier if we could just call a single function and have `pelias/model` figure it out.
- We loop over the OSM tags once to save CPU, this seems ok, however we are looking for the `name` key to set as a primary value (ie. first in the name values list), so there is a conditional IF inside the loop to handle that value differently.

schema/name_osm.js
```diff
@@ -38,14 +35,4 @@ var OSM_NAMING_SCHEMA = {
   // 'sorting_name':     'sorting'
 };
 
-// this property is considered the 'primary name'
```
missinglink, line 41: This was kinda weird and overly fancy, I removed it.

missinglink:
cc/ @SiarheiFedartsou, @orangejulius

SiarheiFedartsou (approved):
👍🏻

util/parseSemicolonDelimitedValues.js
```diff
@@ -0,0 +1,12 @@
+const _ = require('lodash');
+
+// Split multi-value OSM tags into an Array
+// https://wiki.openstreetmap.org/wiki/Talk:Semi-colon_value_separator
+function parseSemicolonDelimitedValues(value) {
+  return (_.isString(value) ? value : '')
+    .split(';')
+    .map(Function.prototype.call, String.prototype.trim)
```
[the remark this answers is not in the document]
missinglink: But then it won't be [the most clever line of javascript](https://blog.bloomca.me/2017/11/08/the-most-clever-line-of-javascript.html) 😢

util/parseSemicolonDelimitedValues.js
```diff
@@ -0,0 +1,12 @@
...
+
+// Split multi-value OSM tags into an Array
+// https://wiki.openstreetmap.org/wiki/Talk:Semi-colon_value_separator
+function parseSemicolonDelimitedValues(value) {
+  return (_.isString(value) ? value : '')
+    .split(';')
+    .map(v => v.trim())
+    .filter(v => v.length);
+}
+
```
missinglink, line 11:
```diff
diff --git a/util/parseSemicolonDelimitedValues.js b/util/parseSemicolonDelimitedValues.js
index 08e21ac..39c123b 100644
--- a/util/parseSemicolonDelimitedValues.js
+++ b/util/parseSemicolonDelimitedValues.js
@@ -5,8 +5,8 @@ const _ = require('lodash');
 function parseSemicolonDelimitedValues(value) {
   return (_.isString(value) ? value : '')
     .split(';')
-    .map(Function.prototype.call, String.prototype.trim)
-    .filter(Boolean);
+    .map(v => v.trim())
+    .filter(v => v.length);
 }

 module.exports = parseSemicolonDelimitedValues;
\ No newline at end of file
```

util/parseSemicolonDelimitedValues.js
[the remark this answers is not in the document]
orangejulius: ahahahah, right ok.  I have a solution for this that I will open in a subsequent PR :)

schema/name_osm.js
orangejulius: Over the years we've learned many things you _can_ do with Javascript, but shouldn't :P

stream/tag_mapper.js
```diff
@@ -35,34 +36,39 @@ module.exports = function(){
...
-          }
+        const langCode = getNameSuffix( key );
+        if( langCode ){
+          const langValues = parseSemicolonDelimitedValues( value );
+          langValues.forEach(( langValue, i ) => {
+            if ( i === 0 ) {
+              doc.setName( langCode, langValue );
+            } else {
+              doc.setNameAlias( langCode, langValue );
+            }
```
orangejulius, line 47: Ah nice this is much cleaner.

stream/tag_mapper.js
```diff
@@ -71,21 +77,22 @@ module.exports = function(){
...
-          _.get(tags, 'nat_name') ||
-          _.get(tags, 'reg_name') ||
-          doc.getName('en');
+        const defaultName = [
+          ...parseSemicolonDelimitedValues(_.get(tags, 'official_name')),
+          ...parseSemicolonDelimitedValues(_.get(tags, 'int_name')),
+          ...parseSemicolonDelimitedValues(_.get(tags, 'nat_name')),
+          ...parseSemicolonDelimitedValues(_.get(tags, 'reg_name')),
+          ...parseSemicolonDelimitedValues(doc.getName('en'))
+        ].filter(Boolean);
```
orangejulius, line 86: More fanciness :P

So just to confirm, this is because you can't use `||` when assigning to `const`, yeah?

Might actually be a use case for `let` in that case, as it's much more clear what the intention is, plus each of these function calls has to always be evaluated, whereas chained `||` will have an early return as soon as there's a truthy value, right?

orangejulius (commented):
Much cleaner after these changes. I had one idea for a further improvement but it's not mandatory at all.

PR #590, opened 2025-02-24
Memorialize The Most Clever Line of Javascript

orangejulius:
It's important to preserve a link to history even if its time has passed 😆


missinglink (approved):
Hah, I don't think this is necessary but it's funny if you'd like to add it.

PR #601, opened 2026-08-01
Remove name:* from taginfo output

pnorman:
As per the taginfo project documentation do not use wildcard characters in any fields. A * is a literal asterisk character.

I don't have the test environment set up to test this change.


orangejulius:
Thanks for the tip! Fixed in https://github.com/pelias/openstreetmap/pull/602

